Issue #77, opened 2012-03-15
No good way to +transition a +transform?


machty:
Hi,

There doesn't seem to be a good way to use bourbon to specify a transition of transforms. For instance, if I want to generate cross-browser CSS to specify a transition for a scale transform, it'd be nice to be able to do something like:

+transition(+transforms)

Obviously, this isn't proper SASS, but ideally it'd generate 

transition: transform
-webkit-transition: -webkit-transform
-moz-transition: -moz-transform
-o-transition: -o-transform

Perhaps the best way to do this would be the following syntax:

+transition($prefixed(transform)) 

That would allow you to combine property specifications such as:

+transition((opacity, $prefixed(transform)), (1.0s, 2.0s) )

PR #86, opened 2012-03-30
Vendor-prefixed *transition-property* names

laurens:
A Solution for issue #77:

Added a helper function to vendor-prefix _transition-property_ names.

The _transition-property_ mixin now replaces the _transform_ property name with appropriate vendor prefixed variants.

No solution for the shorthand _transition_ mixin yet, though. I think this would require string replacement, which seems not to be possible without going to ruby, if I’m seeing it right.

### DIFF
--- a/app/assets/stylesheets/_bourbon.scss
+++ b/app/assets/stylesheets/_bourbon.scss
@@ -7,6 +7,7 @@
 @import "functions/radial-gradient";
 @import "functions/render-gradients";
 @import "functions/tint-shade";
+@import "functions/transition_propery_name";
 
 // CSS3 Mixins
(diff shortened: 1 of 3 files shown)


laicosly:
Am I doing this wrong?  If I transition all, bourbon outputs:
```
// .scss syntax
@include transition(all 500ms $ease-in-out-expo);
// .css output
transition: all 500ms cubic-bezier(1, 0, 0, 1);
```

but if I transition transform bourbon doesn't include the time or ease
```
// .scss syntax
@include transition(transform 500ms $ease-in-out-expo);
// .css output
transition: transform;
```

PR #352, opened 2014-03-13
Improve unpack function

borodean:
`unpack` function is cleaner (thanks https://github.com/thoughtbot/bourbon/pull/180#issuecomment-13439821) and can take the number of values that should exist in the resulting syntax. Thus, it is also used in the `size` mixin.


tysongach:
Closing since this has fallen out of date and the comments above haven’t yet been addressed.

PR #390, opened 2014-04-08
Remove unprefixed properties for the 2009 flexbox spec

malept:
Current browsers support the 2012 spec, so it's highly unlikely that any browser will ever support the unprefixed version of the 2009 spec.


tysongach:
I’m going to close this because Bourbon 5.0 (coming soon) will be removing nearly all prefixes in favor of encouraging use of [Autoprefixer](https://github.com/postcss/autoprefixer).

Issue #361, opened 2014-03-20
Removing `spec` prefix


KittyGiraudel:
Using the standard version of a property shouldn't be an option, hence the `spec` prefix shouldn't even exist in Bourbon.

Thus, I think the prefixer mixin could be rewritten like this:

``` scss
@mixin prefixer($property, $value, $prefixes: 'webkit') {
  @if $prefixes {
    @each $prefix in $prefixes {
      #{"-" + $prefix + "-" + $property}: $value;
    }
  }

  #{$property}: $value;
}
```

Feel free to change the default value for `$prefixes`. Here is an example:

``` scss
*, *:after, *:before {
  @include prefixer(box-sizing, border-box, moz);
}
```

``` css
*, *:after, *:before {
  -moz-box-sizing: border-box,
  box-sizing: border-box;
}
```

Simple and efficient.

PR #428, opened 2014-05-19
Update prefixer mixin and remove spec option

HugoGiraudel:
Regarding #361.

I updated the `prefixer` mixin. I added an extra check in the `calc` mixin. I removed the `spec` prefix altogether.

That's pretty much it. Did I miss something?


kaishin:
I don't think I can handle this soon. Anyone who feels like he can tackle it please do.

tysongach:
**Update:** This is planned for 4.2.0. Thanks for effort, @HugoGiraudel!

Couple of notes:

I’m going to change the `$prefixes` variable name in the settings to `$global-prefixes`. This fixes a bug because `$prefixes` is already used as an argument when calling the `prefixer` mixin:

```scss
$global-prefixes: (
  "webkit": true,
  "moz": true,
  "ms": true,
  "o": true
) !default;
```

And then the `prefixer` mixin will be this:

```scss
@mixin prefixer($property, $value, $prefixes: ()) {
  @each $prefix in $prefixes {
    @if not map-has-key($global-prefixes, $prefix) {
      @warn "Unrecognized prefix: `#{$prefix}`.";
    }

    @if map-get($global-prefixes, $prefix) == true {
      #{"-" + $prefix + "-" + $property}: $value;
    }
  }

  #{$property}: $value;
}
```


Secondly, I’m going to change the `@if` statement in the `prefixer` mixin:

```scss
@if not map-has-key($global-prefixes, $prefix)
```

…to:

```scss
@if map-has-key($global-prefixes, $prefix) == false
```

…for LibSass compatibility.

Let me know if you have any thoughts on these changes.

HugoGiraudel:
```scss
@if map-get($global-prefixes, $prefix) == true {
  #{"-" + $prefix + "-" + $property}: $value;
}
```

Do not use `== true`.

```scss
@if map-has-key($global-prefixes, $prefix) == false
```

LibSass now [supports]() the `not` operator.

tysongach:
Thanks @HugoGiraudel, will do!

Totally forgot LibSass 3.1 implemented the `not` operator.

HugoGiraudel:
Aside of that, it looks nice. Just realised my previous comment was a bit tough, not intended. :P

tysongach:
@HugoGiraudel No worries. Always appreciate the help.

HugoGiraudel:
Anytime for Bourbon, Tyson. ;)

tysongach:
Merged.

Issue #498, opened 2014-08-09
linear-gradient cannot include `transparent`


pathawks:
I am trying to fade from white to transparent, but because bourbon sets the background color to white, the whole background is white.

````sass
@include linear-gradient(white 0px, transparent 100px);
````

This produces the following CSS:

````css
background-color:#fff;
background-image:-webkit-linear-gradient(#fff 0px,transparent 100px);
background-image:linear-gradient(#fff 0px,transparent 100px);
````

The fallback `background-color` prevents any part of the background from being transparent.

bourbon does check to see if the fallback color is transparent, but it only checks the *first* color step.

````sass
  // Set $G1 as the default fallback color
  $fallback-color: nth($G1, 1);

  // If $fallback is a color use that color as the fallback color
  @if (type-of($fallback) == color) or ($fallback == "transparent") {
    $fallback-color: $fallback;
  }
````

**Comment by alanguir** [comment]
You can avoid the $fallback variable by specifying values backwards and using a directional keyword. This lets "transparent" becomes the fallback by virtue of being the first value given:
```
// No gradient, just white
@include linear-gradient(white 0px, transparent 100px);

// Gradient w/ transparency on right, white on left
@include linear-gradient(to left, transparent , white);
```

**Comment by tysongach** [comment]
Don’s comment above should resolve this original issue of fading from `white` to `transparent`:

```scss
@include linear-gradient(white 0px, transparent 100px, $fallback: transparent);
```

Here’s a CodePen: http://codepen.io/tysongach/pen/bdEwvV

It’s also worth noting that unless you need IE9 support, vendor prefixing is [no longer needed](http://caniuse.com/#feat=css-gradients), and you can go with regular ’ol CSS:

```css
background: linear-gradient(180deg, white, transparent);
```

PR #507, opened 2014-08-28
[Issue #498] Added $transparency paramenter to linear-gradient mixin to support layered gradients

Nevraeka:
Added optional `$transparency` variable in the event you want to force transparent `background-color`. `$fallback` will always take the first color stop if it's not set in the mixin. By forcing a transparent background, you can do layered gradient effects.

### DIFF
--- a/app/assets/stylesheets/css3/_linear-gradient.scss
+++ b/app/assets/stylesheets/css3/_linear-gradient.scss
@@ -3,7 +3,8 @@
                        $G5: null, $G6: null,
                        $G7: null, $G8: null,
                        $G9: null, $G10: null,
-                       $fallback: null) {
+                       $fallback: null,
+                       $transparency: false) {
   // Detect what type of value exists in $pos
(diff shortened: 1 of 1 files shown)


joshuaogle:
@Nevraeka Did you try using `$fallback: transparent`? I think that would still solve your use-case unless I this is adding other functionality I don't see.

Nevraeka:
@joshuaogle Yes.  Check out my codepen examples.

tysongach:
I’m not sure I fully understand the use-case here and I want to be sure that that is clear before adding code/new features.

To me, setting `$fallback: transparent` on the `linear-gradient` mixin, as Joshua mentioned above. accomplishes the same thing. Here’s a fork of your CodePen: http://codepen.io/tysongach/pen/aOdmKG

It’s also unclear why layered `div`’s are needed here, as you would be able to accomplish the same thing by using multiple colors and stop within your gradient.

PR #582, opened 2014-11-15
Make _render-gradients and _flex-box become compilable with sass-compiler libraries

thachnn:
- `_flex-box.scss`: `@elseif` is not standard syntax, it should be `@else if`
- `_render-gradients.scss`: should use quote string to concat strings and variables, if not, the sass-compiler will return like `-webkit-linear -gradient(…)`


tysongach:
[Bourbon gem v3.2.4](https://rubygems.org/gems/bourbon/versions/3.2.4) is up.

PR #610, opened 2015-01-03
Add a shorthand-exception-rule argument

schlunz:


schlunz:
# Description:

This addition allows you to provide a sass map

    ( + {top  | right | bottom | left} + : + [space] + {<length> | <percentage> | auto} + )

for the `$shorthand-exception-rule` argument which will be replaced with the already existing set for this CSS property.

# Example:

    [SASS code with addition]

    .class1
        @include position('absolute', 10px, (left: 100px))

    .class2
        @include position(10px, (top: null))

    [SASS code without addition]

    .class1
        @include position('absolute', 10px 10px 10px 100px)

    .class2
        @include position(null 10px 10px 10px)
    

    [CSS result]

    .class1 {
        position: absolute;
        top: 10px;
        right: 10px;
        bottom: 10px;
        left: 100px;
    }

    .class2 {
        position: relative
        right: 10px
        bottom: 10px
        left: 10px
    }

# Works fine with ...

- all existing use cases of this mixin,
- 1 - 4 coordinates given in the `$coordinates` argument or 2 - 4 coordinates given in the `$position` argument (addition only makes sense if 1 or 2 coordinates are give).

tysongach:
Thanks for the PR! But I do think this overcomplicates this add-on. One of Bourbon’s traits is we hope it’s quick and easy to read as a human.

The current implementation is very clear and it follows CSS’s typical shorthand notation:

```scss
@include position(absolute, 10px 10px 10px 100px);
@include position(absolute, null 10px 100px);
```

schlunz:
Ah okay, thats plausible.

Thank you for your feedback @tysongach.

PR #611, opened 2015-01-03
Improved documentation consistency

jamadam:
The change on `html5-input-types.html` mimics latest change on an other document.

Regarding the `selection.html`, the line endings have happened to be CR-LF and I reverted them to LF.


tysongach:
Merged.

Issue #615, opened 2015-01-08
Error on selection mixin when using libsass


jriram:
I get the next error message when importing bourbon 4.1.0:

```
_selection.scss:4: error reading values after @if
```
I am using the following with libsass:

```
$ sassc --version
sassc: 3.0.2-16-gfbdc
libsass: 3.1.0
sass2scss: 1.0.3
```
Is Bourbon 4.1.0 compatible with this version of libsass?

kaishin:
I can confirm this.

kaishin:
Libsass does not support `&` as of `3.1` http://sass-compatibility.github.io/#ampersand_sassscript.

will3:
I get the same error when using gulp-sass version `1.2.4` (https://www.npmjs.com/package/gulp-sass).

andrewcroce:
Confirming this as well

filp:
Also experiencing this :+1:

gitawego:
I can confirm this as well with sass-loader in webpack.

gosukiwi:
Happens to me too, `$ cat Gemfile.lock | grep sass` returns

```
      sass (~> 3.3)
      jekyll-sass-converter (~> 1.0)
      sass (~> 3.2)
      sprockets-sass
    jekyll-sass-converter (1.0.0)
      sass (~> 3.2)
      sass (>= 3.3)
    sass (3.3.8)
    sprockets-sass (1.3.1)
  sass
```

felixbuenemann:
This also breaks sass 3.3 compatibility, it only works with sass 3.4.

tysongach:
Please note that we still [suggest using Bourbon 3.2.3 with LibSass](https://github.com/thoughtbot/bourbon#requirements).

That being said, there’s [already a PR up](https://github.com/thoughtbot/bourbon/pull/616) to refactor the `selection` mixin to work with LibSass.

felixbuenemann:
@tysongach The problem is that the current implementation also throws an exception on the latest version of sass 3.3, while the Gemfile claims to be compatible with sass 3.3+.

javierjulio:
@tysongach, yes @felixbuenemann is right. My Rails 4.2 app now using Bourbon (4.1.1) and sass (3.3.14) and I'm receiving a similar error: `Invalid CSS after " @if ": expected expression (e.g. 1px, bold), was "& {"`

tysongach:
[...]

PR #616, opened 2015-01-08
Make selection() compatible with Libsass

kaishin:
This should address #615

### DIFF
--- a/app/assets/stylesheets/css3/_selection.scss
+++ b/app/assets/stylesheets/css3/_selection.scss
@@ -1,15 +1,42 @@
-@mixin selection {
-  $before-colons: "";
+@charset "UTF-8";
 
-  @if & {
-    $before-colons: "&";
-  }
+/// Outputs the spec and prefixed versions of the `::selection` pseudo-element.
+///
+/// @param {Bool} $current-selector [false]
+///   If set to `true`, it takes the current element into consideration.
+///
+/// @example scss - Usage
+///   .element {
+///     @include selection(true) {
+///       background-color: #ffbb52;
+///     }
+///   }
+///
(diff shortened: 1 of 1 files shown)


joshuaogle:
Why does Libsass have to be such a party pooper?

This is a clearer version of what we had before. Looks great :+1:

felixbuenemann:
This also fixes compatibility with sass 3.3.

gosukiwi:
This is great! Thanks a lot. Hopefully it gets merged soon :)

tysongach:
:+1: 

This breaks current usage when using it on a selector (not a root level), though, so it’d have to wait until Bourbon 5.0.0. Bummer. How do we feel about making an exception and putting it with 4.2.0?

app/assets/stylesheets/css3/_selection.scss
```diff
@@ -1,14 +1,40 @@
...
+    &::selection {
+      @content;
+    }
 
-  #{$before-colons}::selection {
-    @content;
-  }
-  #{$before-colons}::-moz-selection {
-    @content;
+    &::-moz-selection {
```
tysongach: The `-moz-` prefixes should go before the non-prefixed versions.

app/assets/stylesheets/css3/_selection.scss
```diff
@@ -1,14 +1,40 @@
...
+      @content;
+    }
+  } @else {
```
tysongach: Should we have one empty line above this?

javierjulio:
Great to hear this fixes issue with sass 3.3! Thanks @kaishin. Looking forward to the next release.

PR #617, opened 2015-01-09
Fix `font-url` for `sass-rails` > 5.x

### LINKED ISSUE #603: font-face is no longer working with `$asset-pipeline: true`
It is just passing the value in unchanged. Possibly a bug introduced in #595

iwz:
When using newer versions of `sass-rails`, the `$asset-pipeline` flag wasn't
working properly. Although the `font-url` asset helper from
`sass-rails` was being generated into the resulting CSS, it was not being
post-processed by the asset pipeline.

The way we were previously outputting the `font-url` caused `sass-rails` to
not recognize the `font-url` helper. This change appends the helper in a more
direct way, and this resolves the issue.

Fixes #603

### DIFF
--- a/app/assets/stylesheets/helpers/_font-source-declaration.scss
+++ b/app/assets/stylesheets/helpers/_font-source-declaration.scss
@@ -19,16 +19,23 @@
   $src: null;
 
   $formats-map: (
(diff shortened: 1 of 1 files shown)


whmii:
Whew, that was a tough one.

tysongach:
:+1: Works for me.

Thanks @whmii & @iwz!

PR #618, opened 2015-01-09
Setup test suite

tysongach:
* Setup test suite with rspec and css parser
* Add test function for modular-scale

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,9 +1,11 @@
 # encoding: utf-8
-require 'rubygems'
-require 'bundler'
-require 'cucumber/rake/task'
+require "rubygems"
+require "bundler"
+require "cucumber/rake/task"
+require "rspec/core/rake_task"
 
 Bundler::GemHelper.install_tasks
 Cucumber::Rake::Task.new
+RSpec::Core::RakeTask.new(:spec)
 
-task :default => :cucumber
+task :default => [:spec, :cucumber]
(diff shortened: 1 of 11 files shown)


joshuaogle:
Aside from those things that Hound caught, this is looking great. Good job, guys! Really looking forward to having this in.

Those old cucumber tests sure are noisy. I completely missed that it was also running the new ones when you run `rake`. Are we planning on updating those?

joshuaogle:
At what point do we add documentation about tests? Do we want to sneak a few more in first?

bourbon.gemspec
```diff
@@ -28,5 +28,7 @@ meaning they should be as close to the original CSS syntax as possible.
   s.add_dependency('thor')
 
   s.add_development_dependency('aruba', '~> 0.4')
-  s.add_development_dependency('rake')
+  s.add_development_dependency('rake') 
+  s.add_development_dependency('rspec', '~> 2.14.1')
```
teoljungberg: why is `rspec` locked?

Rakefile
```diff
@@ -2,8 +2,10 @@
 require 'rubygems'
 require 'bundler'
 require 'cucumber/rake/task'
+require 'rspec/core/rake_task'
```
teoljungberg: I don't know why hound missed this.. replace with double quotes

donokuda:
Looking good here to me :+1:

PR #619, opened 2015-01-11
Add textarea and button to addons

tysongach:
I find it odd that the [HTML5 Input Types](http://bourbon.io/docs/#html5-input-types) variables for inputs and buttons do *not* include the `textarea` element or `button` element. I can’t think of a use-case where you would not want to also include those. Even in Bitters, we [add `textarea` to the declaration](https://github.com/thoughtbot/bitters/blob/master/app/assets/stylesheets/_forms.scss#L29).

- Less focus on HTML5 inputs and more about targeting *all* text-based inputs, and *all* buttons
- Add `textarea` to `$all-text-inputs`
- Add `button` to `$all-button-inputs`
- Break-out text-based inputs and buttons into separate partials
- Adds SassDoc-style inline documentation

This change, to me, means that variable `#{all-button-inputs}` is no longer clear, because it’s not about inputs. So the variable is now `#{all-buttons}` but I left an assignment for `#{all-button-inputs}` so that this is not a breaking change. Bourbon v5.0.0 can remove the double assignment.

If this change goes into affect, we need to:

- [ ] Update [Bitters](https://github.com/thoughtbot/bitters/blob/master/app/assets/stylesheets/_forms.scss#L29)
- [ ] Update Bourbon’s [current docs](http://bourbon.io/docs/#html5-input-types)


joshuaogle:
These look great! I always assumed there was a reason each of those was missing but after some digging I can't find one, and after some testing/playing around these work fine. :+1:

Also, +1 for renaming and including the aliases.

app/assets/stylesheets/addons/_text-inputs.scss
```diff
@@ -0,0 +1,112 @@
...
+                   'input[type="month"]',
+                   'input[type="number"]',
+                   'input[type="password"]',
+                   'input[type="search"]',
+                   'input[type="tel"]',
+                   'input[type="text"]',
+                   'input[type="time"]',
+                   'input[type="url"]',
+                   'input[type="week"]',
+                   'textarea';
```
tysongach: I know buddy, but we have double quotes inside the single quotes in this instance…

PR #623, opened 2015-01-15
Restore IE8 support in clearfix mixin

matjack1:
Using the single colon pseudo element you allow a better browser support.

This reference: http://caniuse.com/#feat=css-gencontent might be useful.


tysongach:
We don’t support [IE8](https://github.com/thoughtbot/bourbon#browser-support).

PR #626, opened 2015-01-16
Clear up position add-on documentation

tysongach:
- Remove mention of pixels, because who uses those?!
- Remove the unit from the `0`, as our [style guide discourages doing this](https://github.com/thoughtbot/guides/tree/master/style/sass#formatting)


jasonramirez:
@tysongach barring that this actually works in Borbon (ie 0 is parsed as a 0 value and not null) all good! :hamburger:

joshuaogle:
I've been using it this way anyway, didn't realize the docs still had px. Works fine :+1:

tysongach:
Merged.

PR #631, opened 2015-01-17
Changes a word in document

jamadam:
Would you check for this? 
I'm not a native English speaker, though. :)


tysongach:
:+1: Nice catch, thanks!

PR #635, opened 2015-01-21
Add note about Sass's native @import, vs Sprockets directives

cllns:
If someone is converting an existing application that uses CSS to Sass/Bourbon, it might be good to tell them about not using Sprockets directives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,11 +72,7 @@ For command line help, visit our wiki page on Bourbon’s [command line interfac
   mv app/assets/stylesheets/application.css app/assets/stylesheets/application.scss
   ```
 
-4. Delete the Sprockets directive in `application.scss`:
-
-  ```scss
-  *= require_tree .
-  ```
+4. Delete all Sprockets directives in `application.scss` (`require`, `require_tree` and `require_self`), and use Sass's native `@import` instead.
 
 5. Import Bourbon at the beginning of `application.scss`. All additional stylesheets should be imported below Bourbon:
 


cllns:
Apparently I didn't read the next step. Still, I think it could be more clear that the user should not use any Sprockets directives. Right now it's only true if the user has `require_tree` and no other `require`'s
Thoughts?

tysongach:
Instead of adding more inline comments, I think we should just better word #4:

`4. Delete all Sprockets directives in application.scss (require, require_tree and require_self) and use Sass's native @import instead.`

cllns:
Great idea. What do you think about the most recent version, @tysongach?

tysongach:
I think Lines 76–79 are no longer needed and now add confusion.

Also, no need to add “For example” to Line 81.

cllns:
Good point! I like the "For example" to emphasize that it's just an example of what your code could look like. It won't work if the user doesn't have a `home` or `user` stylesheet. 

But I removed it.

tysongach:
I think your previous wording for #4 was more succinct: “…and use Sass’s native `@import` instead…”.

cllns:
Makes sense. I had removed it because it seemed redundant when we had the code block there.

tysongach:
:+1: This is looking really good. Thanks for your help with clarifying this!

Can you squash your commits down to one and I’ll get this merged in?

cllns:
Thanks! all squashed

tysongach:
Merged.

PR #638, opened 2015-01-21
Use prefixer for transition-property

### LINKED ISSUE #637: transition-property doesn't respect global prefixer options
If I set

```
$prefix-for-webkit: false;
$prefix-for-mozilla: false;
```

then

```
@include transition-property(border-color, color);
```

still outputs

```
-webkit-transition-property: border-color, color;
-moz-transition-property: border-color, color;
transition-property: border-color, color;
```

:disappointed:

joshuaogle:
Fixes #637

### DIFF
--- a/app/assets/stylesheets/css3/_transition.scss
+++ b/app/assets/stylesheets/css3/_transition.scss
@@ -58,9 +58,7 @@
 }
 
 @mixin transition-property ($properties...) {
-   -webkit-transition-property: transition-property-names($properties, 'webkit');
-      -moz-transition-property: transition-property-names($properties, 'moz');
-           transition-property: transition-property-names($properties, false);
+  @include prefixer(transition-property, $properties, webkit moz spec);
 }
 
 @mixin transition-duration ($times...) {


app/assets/stylesheets/css3/_transition.scss
```diff
@@ -58,9 +58,7 @@
 }
 
 @mixin transition-property ($properties...) {
-   -webkit-transition-property: transition-property-names($properties, 'webkit');
-      -moz-transition-property: transition-property-names($properties, 'moz');
-           transition-property: transition-property-names($properties, false);
+  @include prefixer(transform-property, $properties, webkit moz spec);
```
tysongach: I think it’d be safe to remove `moz` here.

joshuaogle: I left it in since we are [supposedly supporting Firefox 3.6+](https://github.com/thoughtbot/bourbon#browser-support), and it's only supported the unprefixed version since version 16 (came out in August 2012).

It may be time to update that list though.

tysongach: I almost put up a PR to up the browser support the other day. I'm gonna do that now.

tysongach: [PR](https://github.com/thoughtbot/bourbon/pull/640)

joshuaogle: I was just thinking maybe we should do that at 5.0. That way people are expecting breaking changes anyway.

tysongach: :+1:

tysongach: Also: this should be `transition-property` instead of `transform-property`.

joshuaogle: Whoa, good catch. Fixed.

tysongach:
:+1: LGTM

tysongach:
@joshuaogle This used to call on our `transition-property-names` function. Have you dug into that use? Is that even needed anymore?

joshuaogle:
@tysongach Unfortunately I think it is, I wasn't thinking about needing to prefix property names. And the more I dig into the code the more I think a better rewrite and reorganization is necessary. I'm going to close this PR and put together something more comprehensive

PR #639, opened 2015-01-21
Add a handful of SassDoc comments

tysongach:


app/assets/stylesheets/functions/_shade.scss
```diff
@@ -0,0 +1,24 @@
...
+/// @example scss - Usage
+///   .element {
+///     background-color: shade(#ffbb52, 60%);
+///   }
+///
+/// @example css - CSS Output
+///   .element {
+///     background-color: #664a20;
+///   }
+///
```
joshuaogle: Some of the other examples here just use the naked function for the example rather than including the selector and property. I think I prefer including them, but whichever way we should be consistent. Which do you prefer?

joshuaogle:
Looks good other than what Hound found and the example question (which may be bigger than this PR, so we can tackle that later if you prefer) :rocket:

app/assets/stylesheets/functions/_shade.scss
tysongach: I added a full example declaration here because I imagined people using it in this way. For things like the `contains` function, usually that would only be used within another function or mixin; not in your normal app styles.

joshuaogle: That makes sense. Sounds good then :+1:

PR #640, opened 2015-01-21
Update browser support

tysongach:
Quick update; the triggers:

- Transitions for Chrome, Opera and Safari
- [Box sizing](https://github.com/thoughtbot/bourbon/pull/568) for Firefox

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,11 +102,11 @@ For command line help, visit our wiki page on Bourbon’s [command line interfac
 
 ## Browser support
 
-- Chrome 10.0+
-- Firefox 3.6+
+- Chrome 26+
+- Firefox 29+
 - Internet Explorer 9+
-- Opera 12+
-- Safari 5.1+
+- Opera 15+
+- Safari 6.1+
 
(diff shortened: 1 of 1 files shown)


joshuaogle:
Looking forward to some cleanup for 5.0 with this :+1:

joshuaogle:
How would you feel about bumping Opera to 15 and getting rid of the `-o` prefix? That's when they switched to supporting `-webkit` instead and `-o` is now treated as an alias.

tysongach:
Oh! Nice. Link to source/doc/explanation for that? I’d be game.

joshuaogle:
[This article](http://www.sitepoint.com/opera-css3-webkit-prefix/) talks about the switch and aliases. You can also see the switch on pages like [CanIUse's transform page](http://caniuse.com/#search=transform) if you click "Show all". [Their page on flexbox](http://caniuse.com/#search=flexbox) shows that it became unprefixed not long after, but Safari still requires the `-webkit` prefix anyway so it doesn't matter.

PR #645, opened 2015-01-22
Update docs for font stacks

tysongach:
[Related PR](https://github.com/thoughtbot/bourbon/pull/644)


joshuaogle:
I like it. But can we do a `git rename` instead of just renaming so it maintains history?

tysongach:
@joshuaogle For some reason GitHub is not picking up on `git mv` for me…

joshuaogle:
Oops, I did mean `git mv`. You may need to rename it back the regular way and then do it. I'd like to maintain that, but :+1:

tysongach:
@joshuaogle Man, I tried it all (reverting back and force pushing back here, new commit, rebasing and squashing) but `mv` is just not working for me. It shows as `renamed` locally, but not here on GitHub. :cry:

PR #647, opened 2015-01-22
Break out directional property mixins into partials

tysongach:
The `_directional-values` partial was pretty long and included a bunch of add-ons (e.g. `margin`). This PR breaks them out into individual partials.

I also moved `_border-radius` into the `addons` directory; it’s not a CSS3 thing anymore and it’s already listed under “Add-ons” in our docs.

- Add SassDoc comments
- Some style guide fixes


app/assets/stylesheets/addons/_margin.scss
```diff
@@ -0,0 +1,26 @@
...
+///   .element {
+///     @include margin(null 10px 3em 20vh);
+///   }
+///
+/// @example css - CSS Output
+///   .element {
+///     margin-bottom: 3em;
+///     margin-left: 20vh;
+///     margin-right: 10px;
+///   }
```
joshuaogle: I think I hurt my brain trying to think of how this would actually look

joshuaogle:
Looks good :rocket:

PR #650, opened 2015-01-23
More SassDocs

tysongach:
- Move functions before helpers in main manifest
  - Allows move of `is-number` to the `functions/` directory
- Rename and move is-num helper
  - Aligns with the naming convention of `is-length` and `is-size`
- Add SassDoc for position mixin
- Add SassDoc for is-length function
- Add SassDoc for is-size function
- Add SassDoc for global-prefixes setting
- Rename assign partial to assign-inputs
  - Aligns with our naming conventions


joshuaogle:
:+1:

PR #657, opened 2015-01-29
Revert back to old prefixer

tysongach:
This was a breaking change and we can't offer a fallback. We can update the prefixer when we move to v5.0.0


app/assets/stylesheets/css3/_calc.scss
```diff
@@ -1,6 +1,4 @@
 @mixin calc($property, $value) {
-  @if map-get($prefixes, webkit) {
-    #{$property}: -webkit-calc(#{$value});
-  }
+  #{$property}: -webkit-calc(#{$value});
```
joshuaogle: Oh the irony

joshuaogle:
LGTM :rocket:

PR #659, opened 2015-01-30
Add tests for assign-inputs()

joshuaogle:
Adds a simple test for the `assign-inputs` function.

The last two tests are to make sure it's compiled correctly due to [this bug in Sass](https://github.com/sass/sass/issues/1604), which I believe only affects .sass -> .css conversion, but at least it's covered.

### DIFF
--- a/spec/bourbon/functions/assign_inputs_spec.rb
+++ b/spec/bourbon/functions/assign_inputs_spec.rb
@@ -1,0 +1,50 @@
+require "spec_helper"
+
+describe "assign-inputs" do
+  before(:all) do
+    ParserSupport.parse_file("functions/assign-inputs")
+    @text_inputs_list = [
+      "input[type=\"password\"]",
+      "input[type=\"text\"]",
+      "textarea"
+    ]
+  end
+
+  context "expands plain text inputs" do
+    it "finds selectors" do
+      @text_inputs_list.each do |input|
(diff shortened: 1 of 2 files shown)


spec/bourbon/functions/assign_inputs_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "spec_helper"
+
+describe "assign-inputs" do
+  before(:all) do
+    ParserSupport.parse_file("functions/assign-inputs")
+    @text_inputs_list = [
```
anthonynavarre: Since this variable is dup'ed in most of the tests below, and since we already have some indirection when pulling in the parsed fixture css, I wonder if we could do with a text-based parser so that we can have the CSS definitions that are under spec side-by-side within the test file.

That being said, I don't think it's worth holding up the PR from being merged. What do you think about opening an issue?

anthonynavarre:
Left one suggestion, but overall this LGTM! :+1:

spec/bourbon/functions/assign_inputs_spec.rb
joshuaogle: I do think that's a good suggestion and would be a lot simpler, especially for the smaller tests I have PRs for. This one would be the trickiest, but I'll try to see what it looks like. Since that's a very different question and affects the whole setup, I'll go ahead and merge and put that up separately.

PR #663, opened 2015-01-30
Add test for is-length()

joshuaogle:


spec/bourbon/functions/is_length_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "spec_helper"
+
+describe "is-length" do
+  before(:all) do
+    ParserSupport.parse_file("functions/is-length")
+  end
+
+  context "measures if unitless integer is length" do
```
anthonynavarre: If I understand the purpose of `is-length` correctly, it looks like it is meant to determine whether or not the provided input value can be used as-is to represent a length (requiring both a quantity and units).

Assuming that interpretation is correct (and I'm not sure that it is), can these `context`s be made clearer with something like `"checking if unitless integer can be represented as a length"`?

anthonynavarre:
Left 1 question picking on wording, but implementation looks solid. :+1:

PR #666, opened 2015-02-02
Rename is-light file to match function

joshuaogle:
The function might be less descriptive than the file name, but they should match


tysongach:
Nice catch.

:+1:

PR #668, opened 2015-02-02
Add test for px-to-em()

joshuaogle:
I would like this test to be more robust, but the `em()` function errors when you give it anything other than px anyway.

### DIFF
--- a/spec/bourbon/functions/px_to_em_spec.rb
+++ b/spec/bourbon/functions/px_to_em_spec.rb
@@ -1,0 +1,31 @@
+require "spec_helper"
+
+describe "px-to-em" do
+  before(:all) do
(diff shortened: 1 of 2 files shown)


anthonynavarre:
Per your comment, maybe we can at least add a spec that asserts that we do get a sensible error message if we pass it something other than px (?).

anthonynavarre:
Besides an assertion that an error is raised, are there other cases you already have in mind that would make this test more robust?

joshuaogle:
@anthonynavarre No I don't, because the function is so specific in scope. I'd like to make the function a little smarter, but that would be another PR. For existing functionality, I think the test covers what there is.

anthonynavarre:
@joshuaogle right on. :+1: for what's already here.

PR #669, opened 2015-02-02
When test fails, output css that was found instead

joshuaogle:
Makes testing a little more friendly by showing you more than just pass/fail.


anthonynavarre:
Nice improvement — I dig it. :100:

PR #671, opened 2015-02-06
Style guide fixes

tysongach:


app/assets/stylesheets/addons/_timing-functions.scss
```diff
@@ -6,29 +6,29 @@
...
+$ease-out-circ:     cubic-bezier(0.075, 0.820, 0.165, 1);
+$ease-out-back:     cubic-bezier(0.175, 0.885, 0.320, 1.275);
 
-$ease-in-out-quad:  cubic-bezier(0.455,  0.030, 0.515, 0.955);
-$ease-in-out-cubic: cubic-bezier(0.645,  0.045, 0.355, 1.000);
-$ease-in-out-quart: cubic-bezier(0.770,  0.000, 0.175, 1.000);
-$ease-in-out-quint: cubic-bezier(0.860,  0.000, 0.070, 1.000);
-$ease-in-out-sine:  cubic-bezier(0.445,  0.050, 0.550, 0.950);
-$ease-in-out-expo:  cubic-bezier(1.000,  0.000, 0.000, 1.000);
-$ease-in-out-circ:  cubic-bezier(0.785,  0.135, 0.150, 0.860);
```
joshuaogle: I'm a little torn on this part. It's much easier to make comparisons when they all have the same number of decimals (assuming you know what you are looking at).
We don't have a rule for trailing zeros anyway as far as I know, so maybe it may be preferable in this case.

joshuaogle:
Other than my note on trailing zeros, this is :+1:

app/assets/stylesheets/addons/_timing-functions.scss
tysongach: I totally agree, but I felt that in this case, this Sass will probably never change. These are pretty set in stone; so while we could tell SCSS-Lint to ignore this block, I felt that cleaning it up to match the guide was better (in this case). Thoughts?

joshuaogle: I've used these as reference before, so I assume others do the same. But if it's going to trip up SCSS-Lint, then that's a good enough reason for me :+1:

tysongach: Good point. I just put `_timing-functions.scss` back and added it as an exclude for SCSS-Lint.

PR #685, opened 2015-02-21
Restores app/ for npm installations

kennethormandy:
Right now, installing through npm doesn’t actually include Bourbon. The files were being ignored from when `dist/` was still part of the package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,6 @@
 .gitignore
 .hound.yml
 _site
-app/
 bin/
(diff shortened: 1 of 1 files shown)


pho3nixf1re:
I would love for this to get merged ASAP since it prevents bourbon from being used at all. Anything that needs to be done for this to get in and a version bump for npm?

tysongach:
Merged; thanks! Release coming soon.

PR #692, opened 2015-03-03
Remove redundant specificity from input selectors

whmii:
There has been an ongoing issue in bourbon/bitters where the default selector for input styling is a compound selector (ex. `input[type="email"]`). While this is effective for targeting particular elements, it makes them difficult to override. Now every time you want to style an input element you have to use a compound selector. This PR and a corresponding one in bitter (TBD) removes the unnecessary declaration of the `input` as the element. 

If a user wants to add greater specificity later on, that should be an option and not the default output.

an example…

```scss
input[type="submit"] {
  color: blue;
}

.button {
  color: black;
}
```

` <input type="submit" class="button">` will always render as `color: blue;` even though you would almost always prefer classes to be the dominant selector over inherent element styles.


tysongach:
:+1:

The way this is currently using type + attribute combination selectors does make it overly specific. Class selectors are [more specific](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) than attribute selectors, so removing the type of element like you suggest, allows for authors to override via classes alone.

We should also change Bourbon’s [`#{$all-buttons}`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/addons/_buttons.scss) variable. Can you add those changes to this PR?

I’ll merge for 5.0.

tysongach:
I’ve merged this in for [v5.0.0](https://github.com/thoughtbot/bourbon/tree/bourbon-v5.0.0).

PR #704, opened 2015-03-24
Made clearfix addon an extend instead of an mixin

maryisdead:
The clearfix addon requires no arguments so there's no need to implement it as a mixin. Implementing it as an extend will result in less CSS output (far less, depending on usage count). The proposed change includes a fallback as well.

A more in-depth explanation can be found here: http://blog.teamtreehouse.com/a-better-clearfix-with-sass

### DIFF
--- a/app/assets/stylesheets/addons/_clearfix.scss
+++ b/app/assets/stylesheets/addons/_clearfix.scss
@@ -6,7 +6,7 @@
 ///
 /// @example scss - Usage
 ///   .element {
-///     @include clearfix;
+///     @extend %clearfix;
 ///   }
 ///
 /// @example css - CSS Output
@@ -16,10 +16,15 @@
 ///     display: table;
 ///   }
 
-@mixin clearfix {
+%clearfix {
   &::after {
     clear: both;
     content: "";
     display: table;
   }
+}
+
+// Fallback, deprecate at some point?
+@mixin clearfix {
+  @extend %clearfix
 }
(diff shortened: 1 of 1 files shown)


wardpenney:
When a single declaration is extended more than say, 50 times, I have seen serious performance bugs in the browser. If it is used over 100 times, I've seen it crash Safari. Not to mention that it makes it a giant pain in the Chrome Dev Tools to see where it was coming from. Unless you have source maps on, it is hard to see from where it was extended.

Also, if you extend it from a CSS3 (or higher) declaration, and that declaration is not supported, the browser disregards the entire comma-separated list of selectors for that declaration. It becomes difficult to create fallbacks because they are all coupled together. 

Also, Sass still has severe compilation performance issues when something is `@extend`ed from a deep nest. Not that it is a good idea to nest deeply anyway, but I've seen this exponentially increase dev compile times.

IMO, CSS bloat has never really been a problem. Once gzipped, even the beefiest of CSS files barely breaks 50K. For something of that size, more concerning is the 4095 declaration count max in IE9. 

All in all, in the last couple years, I've used `@extend` less and less. Now I only use it in patterns that are isolated into the same source file. Meaning, their usage is small and is not intended for use across the site. 

Oh and finally, if load-order specificity is important to you, `@extend` renders CSS at the point it was first used. So something you think may be gaining specificity for later load order, may not (by surprise!)

tysongach:
Thanks for the PR! But I agree that the performance increase is negligible and `@extend`’s can more easily lead to problems and abuse.

PR #708, opened 2015-03-29
Add `.sass-cache` to `.npmignore`

killercup:
I just installed bourbon via npm and noticed that the package contained a `.sass-cache` directory. To prevent this from being uploaded in the future, I added it to the `.npmignore` file.


tysongach:
Merged.

Thanks!

PR #713, opened 2015-03-31
Improve tests for variables

joshuaogle:
Tests the output of variables more directly.

### DIFF
--- a/spec/bourbon/addons/buttons_spec.rb
+++ b/spec/bourbon/addons/buttons_spec.rb
@@ -4,49 +4,50 @@ describe "buttons" do
   before(:all) do
     ParserSupport.parse_file("addons/buttons")
 
-    @buttons_list = [
-      "button",
-      "input[type=\"button\"]",
-      "input[type=\"reset\"]",
-      "input[type=\"submit\"]"
-    ]
+    @buttons_list = %w(
+      button
+      input[type="button"]
+      input[type="reset"]
+      input[type="submit"]
+    )
   end
 
   context "expands plain buttons" do
     it "finds selectors" do
-      @buttons_list.each do |input|
-        expect(input).to have_rule("color: #ff0000")
-      end
+      list = @buttons_list.join(", ")
+      ruleset = "content: #{list};"
+
+      expect(".all-buttons").to have_ruleset(ruleset)
     end
   end
 
   context "expands active buttons" do
     it "finds selectors" do
-      list = @buttons_list.dup
-      list.map! { |input| input + ":active" }
-      list.each do |input|
-        expect(input).to have_rule("color: #00ff00")
-      end
+      list = @buttons_list.map { |input| "#{input}:active" }
+      list = list.join(", ")
+      ruleset = "content: #{list};"
+
+      expect(".all-buttons-active").to have_ruleset(ruleset)
     end
   end
 
   context "expands focus buttons" do
     it "finds selectors" do
-      list = @buttons_list.dup
-      list.map! { |input| input + ":focus" }
-      list.each do |input|
-        expect(input).to have_rule("color: #0000ff")
-      end
+      list = @buttons_list.map { |input| "#{input}:focus" }
+      list = list.join(", ")
+      ruleset = "content: #{list};"
+
+      expect(".all-buttons-focus").to have_ruleset(ruleset)
     end
   end
 
   context "expands hover buttons" do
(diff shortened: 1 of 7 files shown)


joshuaogle:
The "has_value" matcher uses id's, so I'm going to leave those. For some reason I had to resort to "has_ruleset" to get it to match when I am mapping the psuedo-classes on to the lists, but I'd like to keep those as id's as well.

spec/bourbon/addons/buttons_spec.rb
```diff
@@ -4,49 +4,50 @@
...
 
   context "expands plain buttons" do
     it "finds selectors" do
-      @buttons_list.each do |input|
-        expect(input).to have_rule("color: #ff0000")
-      end
+      list = @buttons_list.join(", ")
+      ruleset = "value: #{list};"
+
+      expect("#all-button-inputs").to have_ruleset(ruleset)
```
tysongach: While this is still valid (via a fallback), we tweaked the name of this variable to simply `$all-buttons` in https://github.com/thoughtbot/bourbon/commit/a6374b6be375737cdfcb081d3a668b4865692038.

spec/fixtures/addons/buttons.scss
```diff
@@ -1,17 +1,17 @@
 @import "setup";
 
-#{$all-button-inputs} {
-  color: #ff0000;
+#all-button-inputs {
```
tysongach: Can we just go ahead and use class selectors here and for all future tests. Better aligns with how we typically author and Hound will be a happy camper.

spec/spec_helper.rb
```diff
@@ -15,8 +15,4 @@
   config.before(:all) do
     generate_css
   end
-
-  config.after(:all) do
-    clean_up
-  end
```
tysongach: What’s this removal all about?

spec/fixtures/addons/buttons.scss
joshuaogle: I'd have to change the `has_value` matcher, but it should work. Sure

spec/spec_helper.rb
joshuaogle: Doh. That was for testing. Glad you caught that.

spec/fixtures/addons/buttons.scss
tysongach: Ah! I just saw your note below about that. No worries, then. For another time…

spec/fixtures/addons/text-inputs.scss
```diff
@@ -1,17 +1,17 @@
 @import "setup";
 
-#{$all-text-inputs} {
-  color: #ff0000;
+.all-text-inputs {
+  value: $all-text-inputs;
```
tysongach: @joshuaogle Thoughts on disabling Hound on tests?

spec/support/matchers/have_value.rb
```diff
@@ -1,15 +1,17 @@
 RSpec::Matchers.define :have_value do |expected|
   match do |variable|
-    selector_id = variable.sub("$", "#")
-    value_attribute = ParserSupport.parser.find_by_selector(selector_id)[0]
+    selector_id = variable.sub("$", ".")
```
tysongach: Should we rename `selector_id` to `selector_class` now?

spec/fixtures/addons/text-inputs.scss
joshuaogle: :+1: I think there's perfectly valid times like this when bending the styleguide makes for better tests

spec/support/matchers/have_value.rb
joshuaogle: Yep, fixed

spec/fixtures/addons/text-inputs.scss
joshuaogle: On the other hand, we could probably use "content" instead of "value"

tysongach:
This is looking good, @joshuaogle. Thanks for your hard work!

tysongach:
@joshuaogle Is this ready to merge?

joshuaogle:
@tysongach I forgot I still wanted to change those form `value: ` to `content: `. That's done and if you can give it another look over then it should be good to go.

tysongach:
Ah yes, of course! It’s nice not seeing all the red now that `content` is being used. Great work!

LGTM! :+1:

Issue #460, opened 2014-06-23
Import mixins/functions individually


UltCombo:
Hello,

I just started checking this awesome Sass library, but it seems to lack the possibility to import mixins/functions on demand.

Bourbon is deemed as more lightweight than Compass, however being forced to import the whole library just to use a single mixin is a no go for performance -- Bourbon adds around 33% to my build time.

Compass allows importing specific modules, I'd like to see something similar in Bourbon.

For example, I'd expect importing `bourbon/css3/linear-gradient` would make the `linear-gradient` mixin available, but importing it this way does not work as it lacks dependency functions (`_linear-positions-parser`).

<hr>

## Implementation notes

Importing all dependencies in each mixin would re-define the functions, resulting in an even greater performance penalty when importing the whole library, and I'm not sure if it would cause any side effects.

Using the [`import-once`](https://github.com/Compass/compass/blob/master/import-once/README.md) gem would remove the re-importing overhead, but could also generate different output for the user code. I'm not sure if there's a better way to handle the dependency system until this [Sass issue](https://github.com/sass/sass/issues/139) is resolved.

Thoughts?

aeberlin:
This would be great, because these new deprecations are going to drive me insane and force me away from bourbon in the long run.

tysongach:
I’m going to close this because Sass 4 is expected to introduce a native module system and will be much more robust than what we can do on our own.

The system is still being designed, but you can read more about it here: https://github.com/sass/proposal.module-system

PR #719, opened 2015-04-17
Fix 'Undefined variable: "$em-base"'

caseywebdev:
Before this patch:

```bash
$ scss app/assets/stylesheets/functions/_modular-scale.scss
Error: Undefined variable: "$em-base".
        on line 21 of app/assets/stylesheets/functions/_modular-scale.scss
  Use --trace for backtrace.
```

And after:

```bash
$ scss app/assets/stylesheets/functions/_modular-scale.scss
$ echo $?
0
```

### DIFF
--- a/app/assets/stylesheets/functions/_modular-scale.scss
+++ b/app/assets/stylesheets/functions/_modular-scale.scss
@@ -1,3 +1,5 @@
+@import "../settings/px-to-em";
+
 // Scaling Variables
 $golden:           1.618;
 $minor-second:     1.067;


tylerlee:
:+1:

tysongach:
@caseywebdev Hm, that’s strange! Bourbon has one master manifest file that imports all of the partials, including `px-to-em` on [Line 7](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/_bourbon.scss#L7) (this is where `$em-base` is defined). So we shouldn’t need to add this import.

What version of Sass and Bourbon are you using? What does your dev environment look like?

caseywebdev:
I use [cogs](https://github.com/caseywebdev/cogs) for preprocessing, which creates a dependency graph by compiling file dependencies, and those dependencies' dependencies, etc. This allows the creation of a manifest so that subsequent compiles can check the MD5 of the dependency graph (much faster than actually sass compiling) and know immediately whether any work needs to be done. This makes the process very efficient by eliminating duplicate work. This works under the assumption that any individual sass file can compile on it's own because it explicitly declares what it needs to import. This is the only file in bourbon that doesn't adhere to this assumption, and as a result it breaks my build.

tysongach:
Interesting…

I’m confused why this is the only file in Bourbon that this doesn’t work on, because we don’t declare any imports inside each partial, but rather in that manifest file. How is this different from all the other partials in Bourbon, which also have dependencies the same way that the modular scale does?

This changes in this PR would cause (in a typical Sass importing setting) for the `px-to-em` partial to be imported twice.

caseywebdev:
I don't believe any of the other files use variables or mixins that they don't either define or explicitly import like this one does.

tysongach:
@caseywebdev _Lots_ of them do. Nearly all of the CSS3 prefixing mixins ([found here](https://github.com/thoughtbot/bourbon/tree/master/app/assets/stylesheets/css3)) utilize the [`prefixer`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/addons/_prefixer.scss) mixin. The `prefixer` mixin itself then uses a series of variables defined in [`/settings/_prefixer.scss`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/settings/_prefixer.scss), quite similarly to how the `modular-scale` uses `$em-base`.

Many of our features also rely heavily on functions found in the [`functions/`](https://github.com/thoughtbot/bourbon/tree/master/app/assets/stylesheets/functions) directory. For example, the [`size`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/addons/_size.scss) mixin depends on the [`is-size`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/functions/_is-size.scss) function, which depends on the [`is-length`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/functions/_is-length.scss) and [`contains`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/functions/_contains.scss) functions.

In all of Bourbon, no individual partial imports another partial on its own. They are all imported via the master manifest file, [`_bourbon.scss`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/_bourbon.scss).

caseywebdev:
Those successfully compile because the compiler doesn't evaluate those variables inside the mixins at compile time, but I see your point. Personally I like all my files to be individually compilable, if you don't think this change makes sense for bourbon I understand.

tysongach:
The problem is that many features depend on the same mixin/function, causing them to be imported multiple times. So we’d have to have some sort of import utility that checks to see if a mixin or function has already been imported and if so, don’t import it again. Similar to what your cogs tool does, I guess. There’s been some discussion around this [already](https://github.com/thoughtbot/bourbon/issues/460), but we haven’t gotten around to looking into the solution (yet).

Issue #702, opened 2015-03-22
[Announcement] An Update on Bourbon v5.0


tysongach:
We’re working hard on the next major version of Bourbon (v5.0) and we’d like to update you with our goals and progress. You can also follow along on with the [v5.0](https://github.com/thoughtbot/bourbon/tree/bourbon-v5.0.0) branch.

First, we’ve decided to remove browser prefixing features from Bourbon. Our main reasons for doing so are:
- It’s no longer best practice for pre-processors like Sass to handle prefixing. Instead, a post-processor is better suited for this task and allows you to write much cleaner Sass (no more `@include`’s for prefixes!).
- [Autoprefixer](https://github.com/postcss/autoprefixer) is a robust, well-maintained tool to post-process prefixes.
- There’s no easy way for us to decide what browser versions Bourbon should support.
- Autoprefixer ties into the [Can I use](http://caniuse.com) database and allows you to [specify the exact browser support](https://github.com/postcss/autoprefixer#browsers) you need, per project. This is quite powerful, yet painless.

Along with that, we are porting our documentation over to [SassDoc](http://sassdoc.com), so that it’s more comprehensive, consistent and easier to contribute to. And we’re also writing feature tests for Bourbon, so that we can make sure the code is always stable.

We hope everyone is as excited for the future of Bourbon as we are!

– The Bourbon Team

samuelbran:
I respect your decisions on this prefixes thing guys, you're making an awesome job Bourbon. Though I have to say that one of the main reasons I started to use Bourbon was because I was tired to write prefixes by hand and the includes, well, made my life a lot easier. I mean is cool to follow up with the latest techniques and tools but is also cool to have as much as you can in just one tool.

- Sam

hotmeteor:
@samuelbran It definitely takes the onus off of Bourbon and other similar tools to keep up-to-date with the myriad of browser prefixes. I know from time spent on the flexbox portion alone it was a huge task.

If you're not already using Autoprefixer I suggest giving it a whirl. It's really fantastic.

franks921:
:+1: 

Although my reasoning for initially using Bourbon was for the browser prefixes, I do agree and support the decision to remove them. It is time to move forward.

@samuelbran It will be no difficult task to extract those mixins to use in your project while upgrading, but if you are going to spend the time doing that you might aswell use Autoprefixer :)

samuelbran:
[...]

PR #721, opened 2015-04-24
Remove 'webkit' from prefixer mixin include.  Add '-webkit-user-select' ...

pocketbird:
...with value of `$webkit-value`.  Set `$webkit-value` default value to `$value`.

### DIFF
--- a/app/assets/stylesheets/css3/_user-select.scss
+++ b/app/assets/stylesheets/css3/_user-select.scss
@@ -1,3 +1,4 @@
-@mixin user-select($value: none) {
-  @include prefixer(user-select, $value, webkit moz ms spec);
+@mixin user-select($value: none, $webkit-value: $value) {
+  -webkit-user-select: $webkit-value;
+  @include prefixer(user-select, $value, moz ms spec);
 }


tysongach:
Hi!

Why remove `-webkit-` from the `prefixer`? Can you provide an example of how this change resolves an issue your having with the current implementation?

pocketbird:
@tysongach when I use `@include user-select(auto);`, `csslint` fails because only Webkit supports `auto` and the others use `all`, or so I thought.

tysongach:
Since `user-select` is a non-standard spec, it changes a lot and implementation is varied. From [this blog post](http://blogs.msdn.com/b/ie/archive/2012/01/11/controlling-selection-with-css-user-select.aspx), it seems IE10 also takes `auto` as a valid value; or at least it did at one time.

Since there isn’t consistency, I’m not going to merge this. I’d recommend turning off linting for this specific property or use something like Autoprexer to take care for prefixing instead of Bourbon ([we now recommend this](https://github.com/thoughtbot/bourbon/issues/702)).

Issue #723, opened 2015-04-28
Hide-text percentage problem


matheusagcosta:
Guys, we have a problem using percentage in `text-indent` on webkit browsers.

__iOS 8.3 Safari (iPad Mini Retina)__
![img_0011](https://cloud.githubusercontent.com/assets/3741582/7371713/fb689fa8-ed98-11e4-99db-15c092715c9c.PNG)

__Chrome 42.0.2311.90 (OSX Yosemite)__
![screen shot 2015-04-28 at 11 24 11](https://cloud.githubusercontent.com/assets/3741582/7371755/37313158-ed99-11e4-90a4-84c8b64d0fc8.png)

__Chrome 42.0.2311.109 (Android Lollipop)__
![screenshot_2015-04-28-11-27-43](https://cloud.githubusercontent.com/assets/3741582/7371940/44eff102-ed9a-11e4-9ba2-f5cb740dba53.png)
 

Then I changed to `-99999px` and voilá! Problem solved.

Any thoughts?

tysongach:
Can you please provide some more context for us:

- In the above screenshots, what’s your expected outcome?
- How is the Sass implemented to give you the current results?

matheusagcosta:
@tysongach I was trying to align logo at the center.

We are importing boubon family by gemfile.

Here's the result:

https://portal.redepagnet.com/rh/users/sign_in

tysongach:
@matheusagcosta Can we see the Sass source?

matheusagcosta:
@tysongach Sure. Here's the snippet:

```sass

.box {
  display: table;
  margin: 15% auto 0;
  max-width: 300px;
  overflow: hidden;
  width: 90%;

  @include media($tablet) {
    margin-top: 25%;
  }

  @include media($mobile) {
    margin-top: 35%;
  }

  h2 {
    @include hide-text;
    background-image: image-url("rh/logo.svg");
    background-position: center center;
    background-repeat: no-repeat;
    display: block;
    margin-bottom: 20px;
    text-indent: -99999px;
  }

  .button {
    @include appearance(none);
    width: 100%;
  }
}
``````

HugoGiraudel:
Did you make sure that the element has no `padding`? @leaverou and I found a bug on WebKit browsers: https://twitter.com/HugoGiraudel/status/565093151904645120. Chromium ticket: https://code.google.com/p/chromium/issues/detail?id=457146.

tysongach:
[...]

PR #724, opened 2015-04-28
Replace 'hide-text' percentage by px

matheusagcosta:
#723


tysongach:
@matheusagcosta There’s a particular reason we use percentage rather than pixel units: http://www.zeldman.com/2012/03/01/replacing-the-9999px-hack-new-image-replacement/

PR #733, opened 2015-05-16
Sibling count mixin

asross:
Following [these](http://stackoverflow.com/questions/8720931/can-css-detect-the-number-of-children-an-element-has) [posts](http://lea.verou.me/2011/01/styling-children-based-on-their-number-with-css3/), I wrote this mixin for styles specific to the number of children (or multiples of the number of children) of a parent element.

It's had enough applications in the work I've done so far that it seems like it might be generally useful to have in a mixin library like Bourbon.

### DIFF
--- a/app/assets/stylesheets/_bourbon.scss
+++ b/app/assets/stylesheets/_bourbon.scss
@@ -83,5 +83,6 @@
 @import "addons/timing-functions";
 @import "addons/triangle";
 @import "addons/word-wrap";
+@import "addons/sibling-count";
 
 @import "bourbon-deprecated-upcoming";
--- a/app/assets/stylesheets/addons/_sibling-count.scss
+++ b/app/assets/stylesheets/addons/_sibling-count.scss
@@ -13,13 +13,13 @@
 ///
 /// @example scss - Usage
 ///   ul.navbar li {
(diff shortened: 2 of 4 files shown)


tysongach:
Thanks for the PR!

Isn’t this technique quickly becoming obsolete now that we have flexbox [gaining traction and support[(http://caniuse.com/#feat=flexbox)? I find myself using flexbox on projects a lot these days and it’s made for these types of situations.

asross:
That's a good point... especially for things like column width, flexbox is the way to go.

There are still some cases where it's necessary to have styles specific to number of siblings. If you have a three element-per-row data-driven list, and you want to apply a specific style only to the last row (which might have one, two, or three elements), I think something like this is required. More generally, if you want to have styles other than width/height change based on sibling count, then flexbox alone won't do it. So I think it's useful, but it is true that flexbox is a better solution to the most common problem for which you'd be tempted to use this :)

tysongach:
We appreciate the PR, but I’m going to close this for now. I think this mixin is solid, but I think it’s such a rare case that you need something like this, and I don’t want to clutter up Bourbon. I would encourage authors to add it to a project on an as-needed basis.

Thanks again for the PR!

PR #734, opened 2015-05-17
Add input without type to text input list.

clemenshelm:
The `type` attribute is optional and defaults to `"text"`. Therefore an input without `type` should be considered a text input.

### DIFF
--- a/app/assets/stylesheets/addons/_text-inputs.scss
+++ b/app/assets/stylesheets/addons/_text-inputs.scss
@@ -104,6 +104,7 @@ $text-inputs-list: 'input[type="color"]',
                    'input[type="time"]',
                    'input[type="url"]',
                    'input[type="week"]',
+                   'input:not([type])',
                    'textarea';
 
 $all-text-inputs:        assign-inputs($text-inputs-list);
--- a/spec/bourbon/addons/text_inputs_spec.rb
+++ b/spec/bourbon/addons/text_inputs_spec.rb
@@ -19,6 +19,7 @@ describe "text-inputs" do
       input[type="time"]
       input[type="url"]
       input[type="week"]
+      input:not([type])
       textarea
     )
   end


joshuaogle:
The problem with the plain input tag is that it is also used for `type="submit"`, which doesn't need to be inheriting the styles for text fields. I think encouraging people to add `type="text"` is preferable to forcing them to switch all of their submit inputs to buttons.

clemenshelm:
I'm sorry, I can't follow. How can a plain input tag both be used for `type="text"` and `type="submit"`? The HTML5 spec states that

> The missing value default is the Text state.
(http://www.w3.org/html/wg/drafts/html/master/semantics.html#the-input-element)

Or do you mean another HTML version?

The reason why I even needed this fix is that [HTMLCompressor](https://github.com/paolochiodi/htmlcompressor) strips `type="text"` from input fields in order to reduce file size. I used the compressor during the build of http://www.getchillbill.com/ and it broke my UI in production, because all of a sudden my inputs weren't formatted anymore.

So even if people (like me) strive to stick to `type="text"`, 3rd party tools might still break their UI accidentally.

joshuaogle:
Ah, I see. That is weird that they do that.

I always put the type as well, but that's a good solution I think. For some reason I thought that selector would still apply. I put together a demo just to prove it works: http://codepen.io/joshuaogle/pen/aOZYGo

I don't see why this would cause any problems then. :+1: from me

clemenshelm:
Yeah, the selector only matches inputs without any `type` attribute. Thanks, @joshuaogle!

tysongach:
Nice, good call! Thanks for the PR. Merged.

tysongach:
Just had this come up on a real project, and adding `input:not([type])` to the list solved it! Loving this addition!

clemenshelm:
Nice! I'm glad I could help :)

PR #741, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "bourbon",
   "description": "A simple and lightweight mixin library for Sass.",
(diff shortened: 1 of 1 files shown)


tysongach:
Thanks for the PR!

I think we’ll go ahead and keep this in here. It’s a nice reference point to quickly see what version of Bourbon is used and it appears that Bower may actually use this info in the future, so no harm.

kkirsche:
I disagree considering their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #744, opened 2015-06-26
Update SCSS-Lint

tysongach:
- Update to latest version
- Update gem name, as they changed it ([ref](https://github.com/brigade/scss-lint/releases/tag/v0.38.0))


treki:
What does it mean that it failed?

tysongach:
@treki When a PR is submitted, Travis CI runs and makes sure Bourbon’s tests pass, which is this case didn’t happen. We just pushed a fix for that test [here](https://github.com/thoughtbot/bourbon/pull/757).

PR #752, opened 2015-07-17
str-slice(): index of first character in a Sass string starts at 1

robocoder:
http://sass-lang.com/documentation/Sass/Script/Functions.html#str_slice-instance_method

### DIFF
--- a/app/assets/stylesheets/css3/_background-image.scss
+++ b/app/assets/stylesheets/css3/_background-image.scss
@@ -12,8 +12,8 @@
     $spec-image: ();
 
     @if (type-of($image) == string) {
-      $url-str:       str-slice($image, 0, 3);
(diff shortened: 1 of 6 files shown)


tysongach:
@robocoder Is there a specific issue you ran into that prompted you to make this change? I can't seem to reproduce any bugs that this is causing as is…or for that matter, see any change in output by introducing this change.

robocoder:
Ran into it with the php version of the scss compiler. It's been fixed there, but reported it here so you don't rely on undocumented behavior.

tysongach:
Gotcha, thanks!

In some real quick testing, it looks like Sass is smart enough to accept both `0` and `1` to grab the first character.

tysongach:
Merged. Thanks!

PR #757, opened 2015-07-24
Fix failing tests on shade and tint functions

tysongach:


dgalarza:
:ship:

joshuaogle:
Do we know why they started failing? We've had a lot of other Travis issues since it was working last so it's hard to tell where it started.

Tests pass locally now though :+1:

tysongach:
@joshuaogle No, not really. But we’re locking in gem dependencies to try to narrow those possibilities down…

PR #758, opened 2015-07-24
Lock in version for gem dependencies

tysongach:


tysongach:
@dgalarza Mind taking a quick peek at this?

bourbon.gemspec
```diff
@@ -21,12 +21,12 @@ meaning they should be as close to the original CSS syntax as possible.
...
+  s.add_runtime_dependency("sass", "~> 3.4.15")
+  s.add_runtime_dependency("thor", "~> 0.19.1")
 
   s.add_development_dependency("aruba", "~> 0.6.2")
-  s.add_development_dependency("css_parser")
-  s.add_development_dependency("rake")
-  s.add_development_dependency("rspec")
+  s.add_development_dependency("css_parser", "~> 1.3.6")
+  s.add_development_dependency("rake", "~> 10.4.2")
+  s.add_development_dependency("rspec", "~> 3.3.0")
```
dgalarza: Do we need to lock all this down to patch versions? I know for `aruba` specifically we did, but we had specific issues with higher versions. If we can, I'd lock them on the minor version instead.

tysongach: Yeah, great point. I don’t think this is necessary.

dgalarza:
:ship:

Issue #729, opened 2015-05-11
Gradient rendering issue


Doomas:
Hi,

I have following
```
@include linear-gradient(to top, #00ff00, #ff0000);
```
is rendered to
```
background-color: #00ff00;
background-image: -webkit-linear-gradient(bottom, #00ff00, #ff0000);
background-image: linear-gradient(to top,#00ff00, #ff0000, , , , , , , , );
```
The commas at the end is not valid CSS code and therefore the gradient is not displayed in the Browser.

If I Remove the unquote function inside [_linear_gradient.scss] (https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/css3/_linear-gradient.scss) like this 
```
background-image: linear-gradient($pos-spec $full);
```
then it renders correctly to
```
background-color: #00ff00;
background-image: -webkit-linear-gradient(bottom, #00ff00, #ff0000);
background-image: linear-gradient(to top, #00ff00, #ff0000);
```
I'm  not sure if this is a Bourbon problem or a problem with libsass. 

I'm using node-sass 3.0.0 ( libsass 3.2.2 )

tysongach:
Thanks for the heads up, @Doomas.

I can re-create this.

greggmoritz:
I'm also experiencing this on Libsass >3.1.0. Look out for an incoming PR.

georgeescott:
@greggmoritz I too have this issue and have modified the _linear-gradient.scss to match your PR #731 , but it compiles as `linear-gradient(to bottom,#000 #000);` without the comma separating the colour values.

georgeescott:
Working example: @greggmoritz If you update your PR to this it'll work:
```scss
$gradients: $g1, $g2, $g3, $g4, $g5, $g6, $g7, $g8, $g9, $g10;

// Purge null values from list
$full: ();
@each $item in $gradients {
  @if ($item) {
    $full: append($full, $item);
  }
}
$full: zip($full);
```

Then at the bottom change `$full-purged` back to `$full`:
``` scss
background-image: unquote("linear-gradient(#{$pos-spec}#{$full})");
```

mistergraphx:
:+1:

greggmoritz:
@georgeescott Can you provide a little more detail about the engine you're running the compilation on? As well as your simple test case (input, output). I'm unable to reproduce the issue you are seeing. I've run the following sass (taken from [Bourbon docs](http://bourbon.io/docs/#linear-gradient)):

```
[...]

PR #759, opened 2015-07-29
Restore Mozilla Firefox linear gradient support

bookwyrm:


tysongach:
Hi, thanks for the PR!

Is there a particular reason we need to add prefixing back for Firefox? [According to Can I Use](http://caniuse.com/#feat=css-gradients), gradients have been supported there since version 16, which has a global usage of 0.04%.

bookwyrm:
I added it locally to fix a gradient that wasn't displaying in Firefox but now that I take another look, I think it will work if the extra commas in the outputted CSS are removed.

With this SCSS:
```
@include linear-gradient(
  to right,
  color(gradient, #294a9f) 0,
  color(gradient, #294a9f) 10%,
  color(gradient, #3cadac) 90%,
  color(gradient, #3cadac) 100%
);
```

I'm seeing this CSS from Bourbon:
```
background-color: #294a9f;
background-image: -webkit-linear-gradient(left, #294a9f 0, #294a9f 10%, #3cadac 90%, #3cadac 100%);
background-image: linear-gradient(to right,#294a9f 0, #294a9f 10%, #3cadac 90%, #3cadac 100%, , , , , , ); 
```

With that, Firefox uses the fallback color rather than the proper linear-gradient. However, if I manually get rid of the extra commas, it works as expected:
```
background-color: #294a9f;
background-image: -webkit-linear-gradient(left, #294a9f 0, #294a9f 10%, #3cadac 90%, #3cadac 100%);
background-image: linear-gradient(to right,#294a9f 0, #294a9f 10%, #3cadac 90%, #3cadac 100%); 
```

I created a test page with a "bad" (first) and "good" (second) example: http://mattvanderpol.com/gradient-test.html

If you look in latest Chrome/Safari, both gradients render but in latest Firefox only the 2nd ("good") gradient renders.

It just occurred to me that this might be an artifact of my Sass preprocessor (libsass). I'm using `"grunt-sass": "~1.0.0"` which is in turn using `"node-sass": "^3.0.0"`.

tysongach:
Ah gotcha. This seems to be a duplicate of #729, then. Closing this in favor of consolidating the conversation over there.

PR #761, opened 2015-08-09
Make Bourbon a proper npm package

tysongach:
I have no idea what I’m doing…but this does work locally for me. I think this makes Bourbon a proper npm package that can easily be used with things like build tools.

Example `Gulpfile.js`:

```js
var gulp = require('gulp');
var sass = require('gulp-sass');

gulp.task('stylesheets', function () {
  gulp.src('./stylesheets/**/*.scss')
    .pipe(sass({
      includePaths: require('bourbon').includePaths
    }))
    .pipe(gulp.dest('./stylesheets'));
});

gulp.task('default', function () {
  gulp.watch('./stylesheets/**/*.scss', ['stylesheets']);
});
```

@creuter @joshuaogle Do either of you have any context on this stuff? I’d love a review.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,8 @@
 .gitignore
 .hound.yml
 .sass-cache
+.scss-lint.yml
+.travis.yml
 _site
 bin/
 bourbon.gemspec
--- a/index.js
+++ b/index.js
@@ -1,0 +1,7 @@
+var path = require('path');
+
+module.exports = {
+  includePaths: [
+    path.join(__dirname, 'app/assets/stylesheets')
+  ]
+};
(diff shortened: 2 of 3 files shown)


creuter:
I don't have any context, but isn't this already a "proper" npm package? `npm install --save-dev bourbon` worked fine on a recent npm project, I just had to include the whole path with my task runner.

Is this just changing it so you can include `bourbon` rather than `bourbon/app/assets/stylesheets/` in your task runner?

tysongach:
@creuter Yep. With the above gulp config, one can use `@import "bourbon";` in their stylesheets.

wardpenney:
It seems pretty good, and I think it might replace [node-bourbon](https://www.npmjs.com/package/node-bourbon) which I've used before :shipit:

PR #763, opened 2015-08-16
Fix `bourbon update` when Bourbon installed in a custom path

mehlah:
The `update` method removes Bourbon's directory and reinstall it.
The directory removed was wrong in case of a custom installation path.
This commit fix it by making it removing the custom path directory.

### DIFF
--- a/lib/bourbon/generator.rb
+++ b/lib/bourbon/generator.rb
@@ -54,7 +54,7 @@ module Bourbon
     end
 
     def remove_bourbon_directory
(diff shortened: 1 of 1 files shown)


tysongach:
@mehlah Thanks for the PR! Apologies for not responding to you until now.

Our focus right now is to get 5.0 shipped, and once that happens we’ll take a deeper look into this.

dgalarza:
Thanks @mehlah!

PR #764, opened 2015-08-21
Remove unnecessary vendor prefixes from transition

acconrad:
The transition property should only use a polyfill if transitioning the `transform` property.

Transitions by themselves no longer require vendor prefixes ([source 1](https://developer.mozilla.org/en-US/docs/Web/CSS/transition)[source 2](http://caniuse.com/#feat=css-transitions)).

Transitions only require a polyfill if supporting transforms, which only need the `-webkit` prefix. ([source](http://caniuse.com/#feat=transforms2d))

### DIFF
--- a/app/assets/stylesheets/css3/_transition.scss
+++ b/app/assets/stylesheets/css3/_transition.scss
@@ -7,7 +7,6 @@
   // Fix for vendor-prefix transform property
   $needs-prefixes: false;
   $webkit: ();
-  $moz: ();
   $spec: ();
 
   // Create lists for vendor-prefixed transform
@@ -15,7 +14,6 @@
     @if nth($list, 1) == "transform" {
       $needs-prefixes: true;
       $list1: -webkit-transform;
-      $list2: -moz-transform;
       $list3: ();
 
       @each $var in $list {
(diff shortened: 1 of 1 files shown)


tysongach:
Hi, thanks for the PR!

We’re working on Bourbon 5.0 now, [which will no longer have vendor prefixing](https://github.com/thoughtbot/bourbon/issues/702). So we’re planning to keep the prefixes as they are now in 4.x, knowing that hey’ve all been stripped out in 5.0. You can follow along on this branch: [`bourbon-v5.0.0`](https://github.com/thoughtbot/bourbon/tree/bourbon-v5.0.0).

acconrad:
Sounds good! What's the ETA on Bourbon 5 release?

~Adam
@adam_conrad <http://twitter.com/adam_conrad>
http://www.adamconrad.net
Find me on LinkedIn <http://www.linkedin.com/in/acconrad>

tysongach:
@acconrad No real ETA, but we’re actively developing it. Code is in good-shape and I’m going to be releasing an alpha today! The long haul ahead is documentation, which we’ve been porting over to [SassDoc](http://sassdoc.com).

PR #765, opened 2015-08-25
Remove left and repeat for radial-gradient in examples

MarionCraftsmen:
After testing `@include background` on CodePen, I noticed a small mistake in your `radial-gradient` example. Therefore I removed the `left repeat` that shouldn't be at the end of the parameters list.


tysongach:
Thanks @MarionCraftsmen!

PR #768, opened 2015-08-30
Add contrasting-color function

tysongach:
Bourbon has had an [`is-light`](https://github.com/thoughtbot/bourbon/blob/master/app/assets/stylesheets/functions/_is-light.scss) function for awhile now (we used to use it in our now-deprecated `button` mixin). This new `contrasting-color` function builds on `is-light` and allows you to pass it arguments to easily output a light or dark color of your choosing based on the lightness of another color.

So for example, our [default button styles](https://github.com/thoughtbot/bitters/blob/master/app/assets/stylesheets/_buttons.scss) in Bitters could become a bit smarter/more flexible (for brevity, I’m only showing the affected styles below): 

```scss
#{$all-buttons} {
  background-color: $action-color;
  color: contrasting-color($action-color, #000, #fff);
}
```

cc: @whmii

- [x] Needs documentation
- [x] Needs tests

### DIFF
--- a/core/_bourbon.scss
+++ b/core/_bourbon.scss
@@ -32,6 +32,7 @@
 @import "bourbon/addons/border-width";
 @import "bourbon/addons/buttons";
 @import "bourbon/addons/clearfix";
+@import "bourbon/addons/contrast-switch";
 @import "bourbon/addons/ellipsis";
 @import "bourbon/addons/font-stacks";
 @import "bourbon/addons/hide-text";
--- a/core/bourbon/addons/_contrast-switch.scss
+++ b/core/bourbon/addons/_contrast-switch.scss
@@ -1,0 +1,42 @@
+@charset "UTF-8";
+
+/// Switches between two colors based on the lightness of a another color. Great
+/// for building button styles.
+///
+/// @param {color} $base-color
+/// The `color` to evaluate lightness against
+///
+/// @param {color} $light-color [#000]
+/// The `color` to be output when `base-color` is light
+///
+/// @param {color} $dark-color [#fff]
+/// The `color` to be output when `base-color` is dark
+///
+/// @example scss
+/// .first-element {
+///   color: contrast-switch(#bae6e6);
+/// }
+///
+/// .second-element {
+///   $button-color: #2d72d9;
+///   background-color: $button-color;
+///   color: contrast-switch($button-color, #222, #eee);
+/// }
+///
+/// @example css
+/// .first-element {
+///   color: #000;
+/// }
+///
+/// .second-element {
+///   background-color: #2d72d9;
+///   color: #eee;
+/// }
+///
+/// @require {function} is-light
+///
+/// @since 5.0.0
+
+@function contrast-switch($base-color, $light-color: #000, $dark-color: #fff) {
+  @return if(is-light($base-color), $light-color, $dark-color);
+}
--- a/spec/bourbon/addons/contrast_switch_spec.rb
+++ b/spec/bourbon/addons/contrast_switch_spec.rb
@@ -1,0 +1,23 @@
+require "spec_helper"
+
+describe "contrast-switch" do
+  before(:all) do
+    ParserSupport.parse_file("addons/contrast-switch")
+  end
+
+  context "called with a light base color" do
+    it "outputs the dark color" do
+      rule = "color: #000;"
+
+      expect(".contrast-switch-light-base").to have_ruleset(rule)
+    end
+  end
+
+  context "called with a dark base color" do
+    it "outputs the light color" do
+      rule = "color: #eee;"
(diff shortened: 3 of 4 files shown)


tysongach:
Just add `#000` and `#fff` as defaults in https://github.com/thoughtbot/bourbon/commit/6b55fecc95f74905a034ac696b899ec04c1cc03d. Not sure if that’s desired or not…

joshuaogle:
This mixin seems very useful (especially for buttons). The term "contrasting color" is often used similarly to "complementary color", which is not what we are doing here, but I can't think of a better term to use.
Code looks good :+1:

joshuaogle:
Oh and don't forget to add a test :smile:

whmii:
I love this idea and ty for making this a function and not a mixin!

I do feel the syntax of `(color, light value, dark value)` seems a little less overly clear.

aaslordhunter:
I agree with @joshuaogle that the name needs to be more obvious. I assumed from the name that if you gave it blue it would give you orange.

whmii:
this might sound loco but I think some sort of `extreme-color()` (not that name obviously) could be more useful to do something like…

```
if color value ≥ 50% 
 $val: white;
if color value < 50% 
 $val: black;
```

whmii:
this as it stands looks more like a weird work around instead of using `invert($color)`

http://sass-lang.com/documentation/Sass/Script/Functions.html#invert-instance_method

whmii:
maybe `contrast-color()` would be more clear and have it make the text white or black depending.??

whmii:
All the comments!

tysongach:
@whmii Sass’s `invert` is not quite the same (it does exactly what you are thinking of @aaslordhunter). I think Bourbon’s `is-light` function is a bit more refined, taking into account our _perceived_ lightness of a color. Reda wrote a great article about this (https://robots.thoughtbot.com/closer-look-color-lightness) and made a neat tool (http://thoughtbot.github.io/color-lightness-test), too.

Here’s a quick CodePen showing Sass’s `invert` and this `contrasting-color`: http://codepen.io/tysongach/pen/VvvXBJ

Also, Bourbon already has the `is-light` function, which is wrapped by this new function. So the logic for _when_ a color swap should happen is already implemented; this just make it easier to use for things like buttons. It let’s the author choose which two colors to swap depending on lightness, with sensible defaults of `#000` and `#fff`.

Naming-wise, I also considered `contrast-color` and `contrast`. I’m not sure `contrast-color` is much more clear, but at least it’s a smidgen shorter. Maybe a name that’s more about the _swapping/switching_ of colors depending on lightness, rather than making it sounds like it _generates_ a contrasting color. Hmm…

core/bourbon/helpers/_contrasting-color.scss
```diff
@@ -0,0 +1,7 @@
+@function contrasting-color($background-color, $light-color: #000, $dark-color: #fff) {
```
whmii: Arg, this is very nitpicky but I feel like `background-color` is a misnomer. what if this is the `text-color` or a `shadow-color`

tysongach: It might not always be text, though…I could use this function to contrast two shapes with color. I see what you mean with `background-color`, but I chose that because it would be the color _behind_ whatever you’re trying to contrast with. Maybe it should just be `$base-color`?

whmii: or initial color but w/e it's just symantics

whmii:
BTW i found a way this could be slimmed down. I'm putting together a pr about it

core/bourbon/helpers/_contrasting-color.scss
tysongach: Semantics? :wink: 

`initial` is good too!

tysongach:
@whmii Unless it’s completely different, can you just add a commit to this PR?

whmii:
k

whmii:
See http://codepen.io/whmii/pen/wKKjpp

whmii:
doojoon-dealwithit

tysongach:
@whmii Yeah, a sorta ternary operator! Thanks for the improvement!

tysongach:
@joshuaogle @aaslordhunter @whmii What if we named the function `contrast-switch` or `contrast-swap`, instead?

whmii:
@tysongach that sounds better to me. Either one of those works

tysongach:
- Changed the name from `contrasting-color` to `contrast-switch`
- Added documentation
- Added tests

joshuaogle:
Yep, that name works. This is going to be really handy :rocket:

whmii:
yep, super pumped

whmii:
noice!

PR #769, opened 2015-08-31
Added -moz- prefix to calc

ezhmd:
Based on http://caniuse.com/#feat=calc , Firefox 4 to 15 needs -moz- prefix for calc.


joshuaogle:
@ezhmd Bourbon only supports [back to Firefox 29](https://github.com/thoughtbot/bourbon#browser-support), so it isn't necessary here. Firefox 40 just came out, after all. When Bourbon 5 comes out and we're using Autoprefixer, it will add the prefixing on those browsers if they do come up.

PR #773, opened 2015-09-04
Add pad alias to padding mixin

tysongach:
We aim for Bourbon’s features to be human-readable and _clarity_ tends to take precedence over brevity. So a lot of our mixin and function names are not shortened or abbreviated. If you open the stylesheet, someone new to Bourbon should be able to quickly understand what things are doing.

There’s a couple places where we might be able to get away with both clarity and brevity, and one such place is the `padding` mixin…

So I’m tossing up the idea of aliasing our current `padding` mixin:

```scss
@include padding(1rem null);
```

…with a mixin defined as `pad`:

```scss
@include pad(1rem null);
```

1. Thoughts?
- Is this worthwhile?
- Are we opening a can of worms by doing this for one mixin, and not for all?

### DIFF
--- a/core/_bourbon.scss
+++ b/core/_bourbon.scss
@@ -37,6 +37,7 @@
 @import "bourbon/addons/hide-text";
 @import "bourbon/addons/margin";
 @import "bourbon/addons/padding";
+@import "bourbon/addons/pad";
 @import "bourbon/addons/position";
 @import "bourbon/addons/prefixer";
 @import "bourbon/addons/size";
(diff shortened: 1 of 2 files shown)


whmii:
tumblr_lwxyjgdbmz1qlo1lto1_400

smharley:
Would having both available lead to having both in a codebase? I think that would make me sad. I'm not sure we could enforce that very clearly in a style guide or with hound

joshuaogle:
I also think it might be confusing to potentially introduce a mixture of the two just to save 4 characters. Our mixins have usually been named after their property, so I'm in favor of keeping just `padding()`.

tysongach:
Great points! Gonna close for now…

PR #777, opened 2015-09-12
Use a less drastic default scale for modular scale

### LINKED ISSUE #776: Change default scale for `modular-scale` function
Currently the default scale for the `modular-scale` function is Perfect Fourth (1.333). This is a pretty drastic scale and seems unlikely as a default typographic system. 
See http://modularscale.com/?1&em&1.25

I think it would make sense to switch the default value to something that can be more easily controlled out of the box. While it in not a large adjustment, I think Major Third (1.25) would be a better default. 
See http://modularscale.com/?1&em&1.25

screen shot 2015-09-10 at 10 04 19 pm

tysongach:
- Closes #776

### DIFF
--- a/core/bourbon/functions/_modular-scale.scss
+++ b/core/bourbon/functions/_modular-scale.scss
@@ -8,7 +8,7 @@
 /// @param {number} $value [$modular-scale-base (1em)]
 /// The base value the scale starts at
 ///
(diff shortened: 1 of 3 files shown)


joshuaogle:
Much more comfortable :+1:

PR #779, opened 2015-09-12
Grid mixins

whmii:
screen shot 2015-09-12 at 5 56 29 pm

http://codepen.io/whmii/pen/gaaJae


whmii:
:dog: :gun:

tysongach:
@whmii Can you provide some context around why this is a good addition and what purpose it serves? I personally have found baseline grids on the web to be wishful thinking and not worth the effort involved. I also think something like this would be better fit within Neat…

whmii:
While I agree that matrix/baseline grids are not for everyone or every project, having a tool to test alignment that is agnostic to the project's content can be very valuable. Even if it is only for comparing a couple of components to check there alignment, a tool like this would be much preferable to pulling up a mother window and using it as a ruler to check alignment (something I have seen many designers do.

Also I think that it is better served in Bourbon for a number of reasons. Mostly, if it were to be in Neat, that would imply that the primary need for it is in the vertical alignment of objects. I think the best and most typical use case is aligning objects horizontally.

whmii:
Also, I did want to mention that tis is just a tool to be used and it's inclusion is not necessarily tied to  some 'Baseline grids be used all the time/ baseline grids shouldn't be used on the web' sort of statement.

tysongach:
Thanks for the clarification! Personally, I think all grid-related features (both vertical and horizontal) should go inside Neat, rather than Bourbon. It already has the [`visual-grid`](http://thoughtbot.github.io/neat-docs/latest/#visual-grid) and this could go right alongside that.

whmii:
I still feel like that says, "hey user if you want a visual grid to help you design? then add this grid library that doesn't have anything to do with what you are trying to accompish."

no comprende

tysongach:
When a visual grid is needed (especially a baseline grid), it’s likely that a full grid system is already in place in that codebase. And that grid system might or should already have a visual grid built-in (e.g. Susy and Neat both have similar things). Why not just extend Neat’s `visual-grid` with the baseline stuff in here? I know we have a lot of work to do on Neat, but that feels more future-friendly to me. If I cared a lot about baseline grids, I would probably care about horizontal grids, too. I think they should be packaged together.

I also feel that baseline grids aren't used that often on the web and that adding these functions here would see little use. They would be the _only_ grid-related thing in Bourbon.

We also proclaim that Bourbon and Neat go hand-in-hand. If an author turned on these functions, as well as Neat’s visual grid, wouldn't they conflict? Settings would have to be in multiple places, as well, and these functions wouldn't be aware of project settings for Neat.

On top of that all, there are also plenty of JavaScript solutions to do this type of thing, which is another option:

- http://keyes.ie/things/baseliner/
- http://basehold.it

My vote is no.

tysongach:
Closing this for now…we can discuss after 5.0 launch, if needed.

PR #780, opened 2015-09-13
Add pixel-floor function

whmii:
Takes a rem value and rounds it down to the nearest value that is devisible by the number of pixels it represents which is useful for a number of reasons that I don't feel like writing about now.


core/bourbon/functions/_pixel-floor.scss
```diff
@@ -0,0 +1,10 @@
+@charset "UTF-8";
+
+// Takes a rem value and rounds it down to the nearest value that is devisible by the number of pixels it represents
+
+@function pixel-floor($rem, $factor: 16) {
+  $pixel-value: strip-unit($rem) * $factor;
+  $new-pixel-value: floor($pixel-value);
+  $new-rem-value: ($new-pixel-value / $factor) * 1rem;
+  return $new-rem-value;
```
tysongach: Should be `@return`, rather than `return`…

tysongach:
@whmii Can you provide more context around why this is a good addition and what purpose it serves? I’m not sure I see what this does. An example or CodePen/Sassmeister would be fantastic and very helpful! My quick copy over to a [CodePen](http://codepen.io/tysongach/pen/MaKYoP?editors=010) doesn’t seem to work.

whmii:
it should maybe probably work… :sweat_smile:

tysongach:
@whmii Can you add an example and explain how this function helps solve a problem?

whmii:
@tysongach So the reason this exists is `line-height`. In working with moma, we've been doing a lot of baseline alignment stuff. Something I discovered is that browsers do some crazy rounding for line height.

If a browser is given a line height of 18.5 pixels (for example), it will try to align the text baseline to the nearest pixel. in doing so it will also round the line height. If it only rounded each line to the nearest pixel *independently* it wouldn't be that big an issue. Instead it takes the initial value, rounds it down, and then draws the text. that means that the alignment becomes more and more off as the line number increases.

Since this is an unreliable number, it's better to do the calculation client side so you can make typographic assumptions based on a value that you can be sure will be rendered in browser.

tysongach:
I’m not sure I fully understand. I’m especially confused as to why pixel-rounding would be unique to `line-height` only?

`line-height` is a funky property in some ways and I’ve learned to always set it with _no_ unit. This is because when you set `line-height` with a unit, you pass that computed value to any descendants. Here’s two great resources further explaining this:

1. http://meyerweb.com/eric/thoughts/2006/02/08/unitless-line-heights/
2. https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values

Overall, this is such a unique situation and I don’t think we should add this to Bourbon. If it helps with MoMA specifically, just add the function there.

PR #782, opened 2015-09-19
Bad require statement in eyeglass-exports.js

chriseppstein:
eyeglass-exports.js file had a bug in it.


tysongach:
Merged. Thanks @chriseppstein!

PR #801, opened 2015-11-18
Pass Display variable in Ellipsis mixin

haniotis:
Ran into a situation where I didn't want the element to be `display: inline-block` and figured others might have as well.

### DIFF
--- a/app/assets/stylesheets/addons/_ellipsis.scss
+++ b/app/assets/stylesheets/addons/_ellipsis.scss
@@ -20,8 +20,8 @@
 ///     word-wrap: normal;
 ///   }
 
-@mixin ellipsis($width: 100%) {
-  display: inline-block;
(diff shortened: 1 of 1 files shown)


joshuaogle:
I've ended up overriding this property almost half the time, so I think that's a good idea. :+1:

haniotis:
@joshuaogle Cool. What needs to be done for Travis to pass?

joshuaogle:
I think that's a problem with our Travis setup at the moment.

We should add tests for this, but it doesn't need to be in this PR. Should be good to go as far as that goes.

haniotis:
Great! Hopefully others have had this issue and we can get this merged.

tysongach:
Merged as f6e089f597372c48823478de13bed2bef85a913e

Thanks!

PR #803, opened 2015-12-04
Make package.json main the same as in bower.json. Closes #802

### LINKED ISSUE #802: Point package.json main to the same file as bower.json main
I believe those two values should be the same. The effect of the discrepancy I'm observing, is wiredep cannot wireup bourbon's npm package (package.json), while it works fine with bower (bower.json). There might be other effects.

Neat doesn't have this issue - both package.json and bower.json point correctly to the right sass file.

danielsiwiec:


tysongach:
Closing; discussion over on #802

PR #811, opened 2015-12-28
Adds :invalid pseudo-class for $all-text-inputs

### LINKED ISSUE #804: Defining $all-text-inputs-invalid ?
I'm just getting started with Bourbon, so there may be a better way to do this.  But it seems to me useful to define a new all-text-inputs-* variable for styling invalid input.  Along the lines of [Bourbon Text Inputs](http://bourbon.io/docs/#text-inputs), I have done this:

    $all-text-inputs-invalid: assign-inputs($text-inputs-list, invalid);

which enables me to write this:
    
    #{$all-text-inputs-invalid} {
        border: 1px solid #f00;
    }

I was surprised not to find something like this in Bourbon already.  Would it be a useful addition?

tesujimath:
Resolves #804.

### DIFF
--- a/app/assets/stylesheets/addons/_text-inputs.scss
+++ b/app/assets/stylesheets/addons/_text-inputs.scss
@@ -107,7 +107,8 @@ $text-inputs-list: 'input[type="color"]',
                    'input:not([type])',
                    'textarea';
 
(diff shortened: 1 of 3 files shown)


tysongach:
Merged via 8a2a9d23ad8945d8722ac080fe09ce717096566c

Thanks!

PR #812, opened 2015-12-28
Adds :invalid pseudo-class for $all-text-inputs

tesujimath:
Companion pull request for the feature resolving #804.

### DIFF
--- a/_includes/addons/text-inputs.html
+++ b/_includes/addons/text-inputs.html
@@ -18,6 +18,10 @@
 #{$all-text-inputs-active} {
   border: 1px solid #00f;
 }
+
+#{$all-text-inputs-invalid} {
+  box-shadow: 0 0 2px 1px #f00;
+}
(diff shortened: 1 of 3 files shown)


tesujimath:
I see hound is complaining, but my code matches the existing style exactly.  Is this blocking the merge?

joshuaogle:
It's not blocking the merge necessarily. It's acceptable here (to me, anyway).

Looks great! :+1:

joshuaogle:
*Note*: We should add some tests. It doesn't need to be in this PR, but we should get it in ASAP

tysongach:
@tesujimath Thanks for the PR! Since we’re including this new `$all-text-inputs-invalid` variable in our 5.0 release, documentation will be handled by SassDoc via source comments, and no longer via the Jekyll site found in `gh-pages`. Based on that, I’m going to close this PR and the new variable will be indicated in our new documentation site.

PR #813, opened 2016-01-08
Remove unused 2011 flex-direction variable

brianium:
I noticed that the variable `value-2011` was unused while studying the source of the `flex-direction` mixin


tysongach:
@brianium Thanks for the PR! We’ve been working on v5.0, which removes this entire feature (flexbox mixins). So I’ll close this for now since it’s going away and not causing a bug.

PR #819, opened 2016-01-20
Add system font stack

tysongach:
I think there are some great use cases for using system fonts. This adds a robust system font stack as a variable.

It was tough to name this one. `$system` would be a terrible name. So I prefixed this with `font-stack-`; I think if we like this approach, then all of these font stack variable should also be prefixed the same way.

More info: https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/

### DIFF
--- a/core/bourbon/addons/_font-stacks.scss
+++ b/core/bourbon/addons/_font-stacks.scss
@@ -6,14 +6,18 @@
 /// @link goo.gl/Cxb26i
 ////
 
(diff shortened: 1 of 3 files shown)


LkeMitchll:
Good idea. I also think using system fonts can be very robust. You could also suggest that this become the default stack for Bitters `$base-font-family`?

PR #834, opened 2016-01-25
Creation of word-break @mixin

silversonicaxel:
Creation of word-break @mixin

### DIFF
--- a/app/assets/stylesheets/_bourbon.scss
+++ b/app/assets/stylesheets/_bourbon.scss
@@ -63,6 +63,7 @@
 @import "css3/transform";
 @import "css3/transition";
 @import "css3/user-select";
+@import "css3/word-break";
 
 @import "addons/border-color";
 @import "addons/border-radius";
--- a/app/assets/stylesheets/css3/_word-break.scss
+++ b/app/assets/stylesheets/css3/_word-break.scss
@@ -1,0 +1,9 @@
+@mixin wordbreak($wordbreak: normal) {
+  // normal|break-all|keep-all|initial|inherit
+  @include prefixer(word-break, $wordbreak, ms spec);
+
+	@if $wordbreak == break-all {
+		// Non standard for webkit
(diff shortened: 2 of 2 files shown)


tysongach:
Hi @silversonicaxel! Thanks for the PR.

Can you clarify the need for this mixin? Is there a specific prefix you need, for a specific browser?

As far as I know, the [CSS `word-break` property](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) doesn’t require prefixes, with the exception of IE8. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#Browser_compatibility) actually discourages the use of the `-ms-`, though, and [Can I Use](http://caniuse.com/#feat=word-break) shows either no support, partial support (i.e. support for only a portion of the spec), or full support, but no indication that prefixes are needed.

Also, as this time, we are working on Bourbon 5, which removes _all_ prefixing mixins and we are encouraging people to use Autoprefixer, instead. More info: https://github.com/thoughtbot/bourbon/issues/702

silversonicaxel:
Hi @tysongach 
I thought it was also interesting (in a retrospective view) to use it just for a better security in limit scenarios. Plus I added an extra line of code to correct a behaviour for webkit non-standard support. 
My scenario was a combination of hyphens and word-break attributes, so I thought it could have been convenient to have also the word-break mixin. That's it. Thanks for the answer anyway.

tysongach:
Does our [`word-wrap`](http://bourbon.io/docs/#word-wrap) mixin not solve your needs?

I’m going to close this, as we don’t want to build mixins around accounting for bugs. The `-ms-` prefix was—seemingly—never actually required (it was just a synonym to the spec property in IE8 and we also no longer support IE8). The non-standard `break-word` value in WebKit was also a bug, and appears to have been fixed: https://bugs.webkit.org/show_bug.cgi?id=123782

PR #837, opened 2016-01-26
Use a Sass map for Bourbon’s global settings

tysongach:
Using variables for global configuration settings in Bourbon is a bit sloppy and can cause some issues.

One issue is that they are not currently name-spaced. So if you already had—for example—a `$modular-scale-base` variable in your project, or from another library/framework, a collision can occur.

We could solve for this by prefixing each Bourbon setting variable like:

```scss
$bourbon-asset-pipeline: …;
$bourbon-global-font-file-formats: …;
// etc…
```

But that’s a lot of `$bourbon-` repetition…

Instead we can use maps to encapsulate settings in a nice name-spaced manner.

```scss
$_bourbon-defaults: (
  "global-font-file-formats": ttf woff2 woff,
  "modular-scale-base": 1em,
  "modular-scale-ratio": $major-third,
  "rails-asset-pipeline": false,
);
```

And if someone would like to change a default setting, they can just put it in a `$bourbon` map within their project:

```scss
$bourbon: (
  "rails-asset-pipeline": true,
);
```

~~:warning: This is a WIP and doesn’t actually work yet; just looking for high-level feedback for now…~~

#### To-do
- [x] Document
- [x] Test


core/bourbon/settings/_settings.scss
```diff
@@ -0,0 +1,43 @@
...
+  "modular-scale": (
+    "base": 1em,
+    "ratio": $major-third,
+  ),
+  "rails-asset-pipeline": false,
+);
+
+/// User overrides of Bourbon configuration settings
+///
+/// @type map
```
whmii: Is this a sass docs thing? what's goin on here?

whmii:
Wow, bourbon has less settings than I would have assumed. I hear you can wrap `map-get()` in a function to make it look cooler. Maybe you should try that :trollface: 

LGTM :kissing_closed_eyes:

core/bourbon/settings/_settings.scss
tysongach: Yes, these are SassDoc comments. This particular line is indicating that this variable is a map.

tysongach:
@HugoGiraudel Would love your review of this, if you’re up for it!

core/bourbon/settings/_bourbon-get-setting.scss
```diff
@@ -0,0 +1,13 @@
...
+///
+/// @argument {string} $setting
+///
+/// @example scss
+///   _bourbon-get-setting(rails-asset-pipeline)
+///
+/// @access private
+
+@function _bourbon-get-setting($setting) {
+  $settings: map-merge($_bourbon-defaults, $bourbon);
```
HugoGiraudel: I must say I don’t know what is the fastest options between the two following expressions:

```scss
@return map-get(map-merge($_bourbon-defaults, $bourbon), $setting);
```

```scss
@return map-get($bourbon, $setting) or map-get($_bourbon-defaults, $setting);
```

It would be interesting to ask someone who knows or to run a benchmark.

core/bourbon/settings/_settings.scss
```diff
@@ -0,0 +1,33 @@
...
+///   Global base ratio for the `modular-scale` function
+///
+/// @property {bool} rails-asset-pipeline [false]
+///   Enable or disable the `$asset-pipeline` variable for all functions that
+///   accept it.
+///
+/// @access private
+
+$_bourbon-defaults: (
+  "global-font-file-formats": ttf woff2 woff,
```
HugoGiraudel:
```scss
('ttf', 'woff2', 'woff')
```

core/bourbon/settings/_settings.scss
```diff
@@ -0,0 +1,33 @@
...
+  "modular-scale-base": 1em,
+  "modular-scale-ratio": $major-third,
+  "rails-asset-pipeline": false,
+);
+
+/// User overrides of Bourbon configuration settings
+///
+/// @type map
+
+$bourbon: ();
```
HugoGiraudel: Missing `!default`.

HugoGiraudel:
Added a few comments but it looks good otherwise. :)

tysongach:
Thanks @HugoGiraudel!

core/bourbon/settings/_bourbon-get-setting.scss
tysongach: Interesting point. We were discussing performance around this…

I like the clarity of your first option and am going to run with that for the moment, but will try to benchmark sometime.

PR #841, opened 2016-02-03
Refactor clearfix mixin to use block display

whmii:
The presence of `display: table` is part of an accommodation for multiple
vertical margins attempting to collapse in to each other. To preserve more
standard functionality, it is preferred that the object is clearfixed without
effecting margin behavior.

See: http://cssmojo.com/the-very-latest-clearfix-reloaded


whmii:
hmm, didn't see that in the tests, will fix this afternoon.

tysongach:
- [ ] Fix broken tests
- [ ] Update the `@link` to the article posted above

PR #842, opened 2016-02-03
Add the ability to update bourbon settings with the `bourbon-setting()` function

whmii:


tysongach:
@whmii Can you please add a description and an example of what this accomplishes?

Also, it’s tough to track a branch off a branch. Especially since the other branch has an open PR that is still a work-in-progress. Can we close this and converse over on #837?

tysongach:
The work in the other PR wasn’t actually working yet, but just for kicks, I also tried the changes here and can’t seem to get it to work as expected: http://codepen.io/tysongach/pen/EPeKVo

tysongach:
#837 has been updated and merged, and no longer reflects/matches changes made here. The current purpose of the `_bourbon-get-setting` function is to serve one-need: retrieve a discrete global setting. I don’t believe mixing that purpose with re-setting, or updating a setting serves a need we have.

PR #846, opened 2016-02-05
Deprecrate the $weight and $style arguments in the font-face mixin

tysongach:
Now that the `font-face` mixin accepts additional CSS rules via the `@content` directive (235671948ef3a9c343c4391d250082a0373c8d83), it no longer makes sense to have weight and style as arguments, which we never need to manipulate.

Example updated usage:

```scss
@include font-face(
    "calibre",
    "fonts/calibre"
  ) {
  font-style: normal;
  font-weight: 600;
}
```

…which renders to:

```css
@font-face {
  font-family: "calibre";
  src: url("fonts/calibre.woff2") format("woff2"), url("fonts/calibre.woff") format("woff"), url("fonts/calibre.ttf") format("truetype");
  font-style: normal;
  font-weight: 600;
}
```


spec/fixtures/addons/font-face-3.scss
```diff
@@ -3,8 +3,6 @@
 @include font-face(
   "pitch",
   "/fonts/pitch",
-  normal,
-  normal,
-  false,
-  "woff"
+  true,
```
wardpenney: Any reason this switched to `true` for the test?

wardpenney:
Makes sense. In theory, `font-family` could also go this way, but there is less reason anyone would need to customize that Rule. :shipit:

joshuaogle:
:+1: I liked that it was a one-liner, but since those values default to `normal` anyway, it will still be one line most of the time. Looks good.

PR #847, opened 2016-02-05
Paths

tysongach:
- We moved away from `app/assets/stylesheets/` and starting using a `core/` directory in c76215b, but forgot to update the path in `lib/bourbon.rb`
  - This fixes an error that currently exists on `master`, in which you will get an error when trying to import Bourbon: `Error: File to import not found or unreadable: bourbon.`
- This PR is also an attempt to reduce unnecessary nuances to the codebase between Bourbon and Neat. Neat has none of the deletions found in this PR and all seems fine without it.

### DIFF
--- a/lib/bourbon.rb
+++ b/lib/bourbon.rb
@@ -4,10 +4,6 @@ $LOAD_PATH.unshift dir unless $LOAD_PATH.include?(dir)
 
 require "bourbon/generator"
 
-unless defined?(Sass)
-  require 'sass'
-end
-
 module Bourbon
   if defined?(Rails) && defined?(Rails::Engine)
     class Engine < ::Rails::Engine
--- a/lib/bourbon.rb
+++ b/lib/bourbon.rb
@@ -11,7 +11,7 @@ end
 module Bourbon
   if defined?(Rails) && defined?(Rails::Engine)
     class Engine < ::Rails::Engine
-      require 'bourbon/engine'
+      require "bourbon/engine"
     end
 
     module Rails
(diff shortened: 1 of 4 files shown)


lib/bourbon.rb
```diff
@@ -1,17 +1,9 @@
-# CodeKit needs relative paths
```
ckundo: Will this break CodeKit?

lib/bourbon.rb
```diff
@@ -1,17 +1,9 @@
...
-dir = File.dirname(__FILE__)
-$LOAD_PATH.unshift dir unless $LOAD_PATH.include?(dir)
-
 require "bourbon/generator"
 
-unless defined?(Sass)
```
ckundo: Looks good (assuming this is required in the gemspec already)

ckundo:
lgtm!

lib/bourbon.rb
tysongach: @bdkjones Hello! We’re looking to merge these changes in, but we are making an assumption here that this stuff is no longer needed. Our sister library [Neat](https://github.com/thoughtbot/neat/)—which is also part of CodeKit—doesn’t appear to have any of this. If you have a chance, I’d love a quick review, so we don’t bust anything!

bdkjones: Oh, I dumped all the Ruby content of Bourbon a long time ago. The only real reason the gem existed was to copy the Bourbon stylesheet assets into a specified location and to return the version of Bourbon. CodeKit took over those tasks itself so all that's bundled into CodeKit are the stylesheet assets from Bourbon.

bdkjones: Thanks for checking, though!

tysongach: @bdkjones Fantastic! Thank you for the speedy response!

PR #848, opened 2016-02-06
Remove install task

tysongach:
This is an artifact left from the days before the Rails Assets Pipeline. Moving forward, this option of install will no longer be available and we'll only support Rails 3.1+, as of Bourbon 5.0

cc: @ckundo


lib/bourbon.rb
```diff
@@ -1,20 +1,6 @@
...
-          load "tasks/install.rake"
-        end
-      end
-    end
-  else
-    bourbon_path = File.expand_path("../../core", __FILE__)
-    ENV["SASS_PATH"] = [ENV["SASS_PATH"], bourbon_path].compact.join(File::PATH_SEPARATOR)
-  end
+  bourbon_path = File.expand_path("../../core", __FILE__)
+  ENV["SASS_PATH"] = [ENV["SASS_PATH"], bourbon_path].compact.join(File::PATH_SEPARATOR)
```
tysongach: Would love recommendations on how to break this up…

lib/bourbon.rb
```diff
@@ -1,20 +1,6 @@
...
-        rake_tasks do
-          load "tasks/install.rake"
-        end
-      end
-    end
-  else
-    bourbon_path = File.expand_path("../../core", __FILE__)
-    ENV["SASS_PATH"] = [ENV["SASS_PATH"], bourbon_path].compact.join(File::PATH_SEPARATOR)
-  end
+  bourbon_path = File.expand_path("../../core", __FILE__)
```
gabebw: It looks like this was in `module Bourbon` because of the Railtie. Now that it's not, I'd move it out of the module entirely, probably right after the `require`s.

lib/bourbon.rb
gabebw: This could be:

```ruby
ENV["SASS_PATH"] = File.join([ENV["SASS_PATH"], bourbon_path].compact)
```

`File.join` is the same as `.join(File::PATH_SEPARATOR)`, but looks nicer and is shorter.

lib/bourbon.rb
gabebw: In fact, this means you could remove the `module Bourbon` in this file entirely since this is the only thing in it.

tute:
> we'll only support Rails 3.1+, as of Bourbon 5.0

We could support Rails 4.2+, as others are EOL'd. If it's not hard and you'd like we can keep supporting older versions too anyway.

lib/bourbon.rb
```diff
@@ -1,20 +1,6 @@
 require "bourbon/generator"
+bourbon_path = File.expand_path("../../core", __FILE__)
```
tute: We can move this line inside of the module.

tute:
We have to fix CI. Code LGTM!

README.md
```diff
@@ -52,7 +52,7 @@ For command line help, visit our wiki page on Bourbon’s [command line interfac
 
   It’s not recommended to add or modify the Bourbon files so that you can update them easily.
```
gabebw: > It’s not recommended to add or modify the Bourbon files so that you can update them easily.

What does that mean? Is this left over from when `bourbon install` was a thing?

gabebw: Oh it is! The `bourbon install` instructions above should go away, right?

tysongach: No, because we still allow people to install the Bourbon gem to their machine, and then just use our CLI to install a local instance of Bourbon into a project. This is for people who may not be in a Rails project, or a project with npm…but maybe just a simple website with no package manager at all.

gabebw: Aha, I confused `rake bourbon:install` with `bourbon install`, which [still exists](https://github.com/thoughtbot/bourbon/blob/34fe6c7897974598ba8a8c28d5bd25f6150588cb/lib/bourbon/generator.rb#L11).

tysongach: :+1:

gabebw:
:+1:

lib/bourbon.rb
```diff
@@ -1,20 +1,3 @@
...
-          load "tasks/install.rake"
-        end
-      end
-    end
-  else
-    bourbon_path = File.expand_path("../../core", __FILE__)
-    ENV["SASS_PATH"] = [ENV["SASS_PATH"], bourbon_path].compact.join(File::PATH_SEPARATOR)
-  end
-end
+bourbon_path = File.expand_path("../../core", __FILE__)
```
tysongach: @gabebw @tute Can `"../../core"` just be `"./core"`?

gabebw: I think it cannot -- the relative path + __FILE__ wonkiness that's necessary to get an absolute path means that `./core` would probably resolve to `bourbonlib/core`, not `bourbon/core`.

PR #860, opened 2016-02-28
`contrast-switch` Updates

tysongach:
- Change `contrast-switch` argument names
  - This makes the name of the argument in line with the color it is outputting, rather than the color of the base, which didn't make sense.
- Add global settings for `contrast-switch` colors

cc: @whmii @ethangl


whmii:
During our discussions last Friday, I was wondering if this shouldn't throw an error if you try to pass a light color in the `dark-color` setting and vise versa. 

Either way, good wok. LGTM

PR #861, opened 2016-03-01
Reorganize core files

tysongach:
- ~~`support`: library set up~~ `helpers`: sort of like “constants” for the library
- `settings`: library settings
- `validators`: simple functions to validate things
- `utilities`: help build features of the core library
- `library`: core user-facing library


joshuaogle:
`settings` and `support` seem awfully similar and each have only one file. I think it would make sense to move `scale.scss` into `settings`. Other than that I like having it split up by purpose and the names look great :+1:

tysongach:
@joshuaogle Yeah, I’m not super happy with `settings` and `support`. The reason I broke them out is because `settings` _uses_ the stuff in `scales`… :confused:

joshuaogle:
@tysongach There's nothing wrong with importing the files in a specific order. Or at least, I don't see a difference between that and importing folders in a specific order. Either way is fine with me, just trying to keep it simple.

joshuaogle:
Yep, that looks better to me too. LGTM :+1:

tysongach:
Thank you, @joshuaogle!

PR #863, opened 2016-03-03
Add the (refactored) triangle mixin back

tysongach:
- Remove the `inset-up`,  `inset-right`, `inset-down`, and `inset-left` (introduced in https://github.com/thoughtbot/bourbon/pull/291, but was never added to the public documentation)
- Remove the ability to pass in a `foreground-color` _and_ `background-color` (also introduced in #291, not sure why) and instead only have one color argument
- Add argument defaults
- Add documentation
- Simplify the Sass
- Provide helpful errors for input on the $direction argument

```scss
@include triangle(up, $color, 1rem);
```

Try it out here: http://codepen.io/tysongach/pen/yONPgJ

### DIFF
--- a/core/_bourbon.scss
+++ b/core/_bourbon.scss
@@ -44,4 +44,5 @@
 @import "bourbon/library/strip-unit";
 @import "bourbon/library/text-inputs";
 @import "bourbon/library/tint";
+@import "bourbon/library/triangle";
 @import "bourbon/library/word-wrap";
--- a/core/bourbon/library/_triangle.scss
+++ b/core/bourbon/library/_triangle.scss
@@ -1,0 +1,79 @@
+@charset "UTF-8";
+
+/// Generates a triangle pointing in a specified direction.
+///
+/// @argument {string} $direction [up]
+///   The direction the triangle should point. Accepts `up`, `up-right`,
+///   `right`, `down-right`, `down`, `down-left`, `left` or `up-left`.
+///
+/// @argument {color} $color [currentColor]
+///   Color of the triangle.
+///
+/// @argument {number (with unit)} $width [1rem]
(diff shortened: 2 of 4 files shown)


tysongach:
For reference, here’s the current one: https://github.com/thoughtbot/bourbon/blob/v4-stable/app/assets/stylesheets/addons/_triangle.scss

kaishin:
What's the reasoning behind re-adding this?

tysongach:
@kaishin I initially removed it because I rarely heard of people using it. Then when I was talking about how it was removed for 5.0, I heard reactions from a few people that they did in fact use it and found it to be a handy resource.

At the end of the day, it’s a pretty lightweight mixin and doesn’t have any dependencies.

jasonramirez:
The code looks clean as a whistle! I didn't pull it and try it locally - do you feel its worth adding tests for this? Code is A+

kaishin:
@tysongach Fair enough. I used it in the past but I personally think it's too specific. Like, we have triangle but not square and circle ;)

Anyways, feel free to merge.

tysongach:
@kaishin Good point. Though squares and circles are very approachable from a code perspective, especially with our `size()` mixin:

```scss
.square {
  @include size(1rem);
  background-color: blue;
}

.circle {
  @include size(1rem);
  background-color: blue;
  border-radius: 50%;
}
```

Whereas with triangles, we use a “hack” with borders to accomplish it, and it takes more effort to think which borders need to be a certain size, which ones needs to be a certain color or transparent, etc.:

```scss
.triangle {
  border-color: #b25c9c transparent transparent;
  border-style: solid;
  border-width: 0.5rem 0.5rem 0;
  height: 0;
  width: 0;
}
```

PR #865, opened 2016-03-04
Add pathname requirement to resolve install issues

dgalarza:
The bourbon CLI depends on `Pathname`, however it is not actually being
required anywhere. The test suite presumably passes because it pathname
is required by a test dependency.


tysongach:
:+1: Thank you, @dgalarza!

PR #878, opened 2016-03-18
Re-order font-face mixin arguments

tysongach:
Swap `$file-formats` and `$asset-pipeline` so that `$asset-pipeline` is last because it is likely used the least.

This means you would be able to write it a bit more succinctly:

```diff
@include font-face(
  "source-sans-pro",
  "fonts/source-sans-pro-regular",
- $file-formats: "woff2"
+ "woff2"
);
```


joshuaogle:
:+1: for putting arguments with defaults at the end. We should go through the error messages at some point and make sure they're set up to help people transition when their old arguments are in the wrong order

PR #882, opened 2016-03-29
Add composer.json for convenient use with PHP apps

shnhrrsn:
This allows PHP based apps to require Bourbon conveniently via their composer file and use it as such: `scss --load-path vendor/thoughtbot/bourbon/core some-stylesheet.scss`.

For more info on composer: http://getcomposer.org/

If this PR is approved, Bourbon should be submitted to https://packagist.org.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,6 +7,7 @@ bin/
 bourbon.gemspec
 bower.json
 circle.yml
+composer.json
 features/
 Gemfile
 Gemfile.lock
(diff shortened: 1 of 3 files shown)


tysongach:
Thanks for the PR, @shnhrrsn!

At this time, we’re not interested in officially supporting Composer or PHP-based package managers. It’s rare for us to work with PHP at thoughtbot and we want to make sure if we are going to support something, that we have the knowledge and experience to do it right.

We’ve had people from the community wrap Bourbon and provide installation via other package managers in the past, so if that is something you’d be interested in, we’d welcome that!

PR #891, opened 2016-04-15
Add deprecation warnings

tysongach:
Introduce remaining deprecation warnings for features that are removed in 5.0.0.

This borrows heavily from how we deprecate features in Neat using Sass `@warn`: thoughtbot/neat@5c337cc

One key difference from Neat is that we only want to use this mixin for
deprecations, not _all_ warnings. The reason is because with CSS, it's not always
a straightforward task to upgrade libraries or frameworks; particularly major
releases. So we offer a toggle to globally disable deprecation warnings, but for
other meaningful warnings we need to throw, we still output those.

People can disable deprecation warnings by setting a variable to `false`:

```scss
$disable-bourbon-deprecation-warnings: true;
```

One problem with this system is that we cannot use these deprecation mixins to
deprecate a function. This is because Sass doesn't allow you to include a mixin
in a function. For now, since we only have a few functions up for deprecation,
we can live with this and manually write warnings for those.

**The plan is to merge this and release it as a `4.3` release to solely introduce
these warnings.**

### DIFF
--- a/app/assets/stylesheets/_bourbon-deprecate.scss
+++ b/app/assets/stylesheets/_bourbon-deprecate.scss
@@ -8,12 +8,12 @@
 @mixin _bourbon-deprecate($feature, $message: null) {
   @if $output-bourbon-deprecation-warnings == true {
     @warn "[Bourbon] [Deprecation] `#{$feature}` is deprecated and will be " +
-      "removed in 5.0.0. #{$message}.";
+      "removed in 5.0.0. #{$message}";
     @content;
   }
 }
 
 @mixin _bourbon-deprecate-for-prefixing($feature) {
   @include _bourbon-deprecate($feature, "We suggest using an automated " +
-    "prefixing tool, like Autoprefixer");
+    "prefixing tool, like Autoprefixer.");
 }
--- a/app/assets/stylesheets/_bourbon-deprecate.scss
+++ b/app/assets/stylesheets/_bourbon-deprecate.scss
@@ -1,0 +1,19 @@
+@charset "UTF-8";
+
+/// Throws Sass warnings to announce library deprecations. You can disable them
+/// by setting the `$output-bourbon-deprecation-warnings` variable to `false`.
+///
+/// @access private
+
+@mixin _bourbon-deprecate($feature, $message: null) {
+  @if $output-bourbon-deprecation-warnings == true {
+    @warn "[Bourbon] [Deprecation] `#{$feature}` is deprecated and will be " +
+      "removed in 5.0.0. #{$message}.";
+    @content;
+  }
+}
+
+@mixin _bourbon-deprecate-for-prefixing($feature) {
+  @include _bourbon-deprecate($feature, "We suggest using an automated " +
+    "prefixing tool, like Autoprefixer");
+}
--- a/app/assets/stylesheets/_bourbon-deprecated-upcoming.scss
+++ b/app/assets/stylesheets/_bourbon-deprecated-upcoming.scss
@@ -1,12 +1,13 @@
 // The following features have been deprecated and will be removed in the next MAJOR version release
 
 @mixin inline-block {
-  display: inline-block;
+  @include _bourbon-deprecate("inline-block");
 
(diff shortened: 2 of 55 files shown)


app/assets/stylesheets/_bourbon-deprecate.scss
```diff
@@ -0,0 +1,19 @@
+@charset "UTF-8";
+
+/// Throws Sass warnings to announce library deprecations. People can disable
+/// these using the `$disable-bourbon-deprecation-warnings` variable.
```
jasonramirez: How about:

```
/// these by setting the `$disable-bourbon-deprecation-warnings` variable to `false`.
```

jasonramirez: There is also something funny to me about the double negative here. What if we were to name the variable `show-bourbon-deprecation-warnings` or `output-bourbon-deprecation-warnings` or something that will get us to a place where we are more direct with `false` turning it off, and `true` turning it on.

app/assets/stylesheets/_bourbon-deprecated-upcoming.scss
```diff
@@ -377,35 +376,50 @@
 
 // Flexible grid
 @function flex-grid($columns, $container-columns: $fg-max-columns) {
+  @if $disable-bourbon-deprecation-warnings == false {
```
jasonramirez: Ugh. Is this because you can't nest functions?

app/assets/stylesheets/css3/_flex-box.scss
```diff
@@ -2,58 +2,78 @@
...
   display: -webkit-box;
   display: -moz-box;
   display: -ms-flexbox; // IE 10
   display: box;
 }
 
 @mixin box-orient($orient: inline-axis) {
+  @include _bourbon-deprecate-for-prefixing("box-orient");
+
 // horizontal|vertical|inline-axis|block-axis|inherit
```
jasonramirez: I know this isn't part of the PR ... but indent?

app/assets/stylesheets/css3/_flex-box.scss
```diff
@@ -2,58 +2,78 @@
...
   @include prefixer(box-orient, $orient, webkit moz spec);
 }
 
 @mixin box-pack($pack: start) {
+  @include _bourbon-deprecate-for-prefixing("box-pack");
+
 // start|end|center|justify
```
jasonramirez: Same as above. Feel free to ignore.

jasonramirez:
Whew that was a long one. Code looks good. I have a particular comment about the double negative variable `$disable-bourbon-deprecation-warnings` and perhaps turning into an explicit positive. Other than that 🍔

app/assets/stylesheets/_bourbon-deprecate.scss
tysongach: Yes, great point. I debated this because this current language aligns with Neat’s implementation (suite consistency), but the double negative is indeed a bummer. `output-bourbon-deprecation-warnings` is much more clear and straightforward.

app/assets/stylesheets/_bourbon-deprecated-upcoming.scss
tysongach: Yeah, can’t put a mixin inside a function.

app/assets/stylesheets/css3/_flex-box.scss
tysongach: Yes, these little inconsistencies drive me bonkers. But since this is all old code that has already been removed in 5.0, I try and not touch it so to create clean diffs and commits, and also not accidentally introduce bugs.

5.0 is super duper consistent and clean, though 😉

tysongach:
Thanks, @jasonramirez!

avit:
In case anyone else got confused, the pull request description says:

```
$disable-bourbon-deprecation-warnings: true;
```

But the actual code has:

```
$output-bourbon-deprecation-warnings: true;
```

PR #895, opened 2016-05-13
Added value-prefixer

matthewtobiasz:
There are some non-standardized value prefixes (e.g., https://developer.mozilla.org/en-US/docs/Web/CSS/cursor ) that require a vendor prefix on the value to get working. I added a utility mixin to output those prefixes on values, very much like the existing prefixer mixin.

### DIFF
--- a/core/_bourbon.scss
+++ b/core/_bourbon.scss
@@ -47,4 +47,5 @@
 @import "bourbon/library/text-inputs";
 @import "bourbon/library/tint";
 @import "bourbon/library/triangle";
+@import "bourbon/library/value-prefixer";
 @import "bourbon/library/word-wrap";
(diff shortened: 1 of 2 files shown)


tysongach:
@matthewtobiasz Thanks for the PR! Having this PR seems like a good idea since we offer one for the property side of prefixing. Thanks for aligning to that.

tysongach:
Couples thoughts:

1. With 5.0 removing all of the prefixing features, we kept the `prefixer()` mixin, so that at least people could prefix a one-off property if they really needed to. It’s a clutch.
1. With most browsers moving towards runtime flags rather than prefixes, `prefixer()` & `value-prefixer()` will probably no longer be needed. We can deprecate when that is a known.

tysongach:
### Todo
- [ ] Add tests

tysongach:
Thanks, @matthewtobiasz! Merged as 0cbaa7b7e589b38f2c2f56a6eca42553b2180155.

PR #896, opened 2016-05-17
Remove word-break from the word-wrap mixin

pat270:
The word-wrap mixin is just a wrapper for `word-break: break-all`. It doesn't make sense to have a word-wrap mixin when you can use word-break: break-all. See https://jsfiddle.net/0z7a9xtk/. I'm assuming the conditional statement with word-break: break-all was added to make it work in tables, but I think that should be left to the user to sort out.

### DIFF
--- a/core/bourbon/library/_word-wrap.scss
+++ b/core/bourbon/library/_word-wrap.scss
@@ -13,17 +13,10 @@
 /// @example css
 ///   .wrapper {
 ///     overflow-wrap: break-word;
-///     word-break: break-all;
 ///     word-wrap: break-word;
 ///   }
 
 @mixin word-wrap($wrap: break-word) {
   overflow-wrap: $wrap;
   word-wrap: $wrap;
-
-  @if $wrap == break-word {
-    word-break: break-all;
-  } @else {
-    word-break: $wrap;
-  }
 }
--- a/spec/bourbon/library/word_wrap_spec.rb
+++ b/spec/bourbon/library/word_wrap_spec.rb
@@ -9,8 +9,7 @@ describe "word-wrap" do
     it "adds word-wrap" do
       input = ".word-wrap"
       ruleset = "overflow-wrap: break-word; " +
-                "word-wrap: break-word; " +
-                "word-break: break-all;"
+                "word-wrap: break-word;"
 
       expect(input).to have_ruleset(ruleset)
     end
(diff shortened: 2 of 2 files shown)


tysongach:
Hi @pat270, thanks for the PR!

I did some digging and have some thoughts:

1. For reference, the `word-wrap` mixin was originally introduced in ddaf365f8957907c370cb283cfc92432cb3673c1 (PR #373).
2. The use of the conditional statement changed over time, but it’s current iteration exists so you can reset the values via the same mixin (by passing the argument `word-wrap(normal)`); this was introduced in 9d8e7fbd62c6d027ac9a6e57066ef700e0a1ded4 (PR #576).
3. CSS’s `overflow-wrap`/`word-wrap` and `word-break` properties serve slightly different purposes:
  - [The spec for `overflow-wrap` (or the legacy name, `word-wrap`) says](https://www.w3.org/TR/css-text-3/#overflow-wrap) that it “specifies whether the UA may arbitrarily break within a word to prevent overflow when an otherwise-unbreakable string is too long to fit within the line box.”
  - [The spec for `word-break` says](https://www.w3.org/TR/css-text-3/#word-break-property) that it “specifies soft wrap opportunities between letters, i.e. where it is ‘normal’ and permissible to break lines of text.” It also seems to indicate that this property is particularly useful for CJK typesetting.
4. I not sure if we need both `overflow-wrap` and `word-break`. On one hand, covering many cases is nice, but on the other, we do not aim to solve for CJK typesetting.
5. The current description of our mixin does a poor job explaining these nuances and we should update it.
6. Maybe we need to rename this mixin to `overflow-wrap` since that is the W3C’s new name for `word-wrap` going forward…


cc: @joshuaogle, who originally authored the mixin.

pat270:
Hey @tysongach, Thanks for the detailed clarification! Did you want an updated pull with a new mixin overflow-wrap and deprecate word-wrap or are you guys still deciding on the solution?

certainlyakey:
This is a very useful and needed PR.

The essential difference between `overflow/word-wrap` and `word-break` is best described here: http://stackoverflow.com/a/15137272/102397. Basically, the first one will break only words longer than the width of container, while the second will break any word **everywhere**, making the text hardly legible because it will turn into a mess with even very short words broken. 

The current `word-wrap` mixin is really sort of useless, because it will either turn on `word-break` on, creating a mess of words cut in half (default argument), or turn both `overflow/word-wrap` and `word-break` off (argument `normal`).

tysongach:
Merged as aa4be0d6a51147c36bf40081db97413d554703f7.

Thank you @pat270 & @certainlyakey!

pat270:
Nice! Thanks @tysongach.

PR #898, opened 2016-05-25
Update Prefixes

vlw0052:
Update prefixes needed for css3 3d transform.


tysongach:
@vlw0052 Thanks for the PR! We’ve shifted focus to our next major release—5.0—which [removes _all_ prefixing features](https://github.com/thoughtbot/bourbon/issues/702).

Since this change is only removing a few prefixes and not patching a bug, I’m going to close the PR.

PR #902, opened 2016-06-01
Add is-color validator

tysongach:
This function allows us to check user input for CSS colors when we expect one. If someone accidentally inputs a non-color, they’ll get a nice error message:

```
Error: `test` is not a valid color for the `$base-color` argument in the `contrast-switch` function.
```

To kick things off, I’ve added this validator to the `contrast-switch`, `shade` & `tint` functions, as well as the `triangle` mixin.

### DIFF
--- a/core/_bourbon.scss
+++ b/core/_bourbon.scss
@@ -12,6 +12,7 @@
 
 @import "bourbon/validators/contains";
 @import "bourbon/validators/contains-falsy";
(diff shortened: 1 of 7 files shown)


joshuaogle:
Took it for a test drive and everything looks great. 💯

tysongach:
Thanks, @joshuaogle!

PR #903, opened 2016-06-01
Fix else if statement

tysongach:
According to the Sass docs, a space should be placed between the two.

Ref: http://sass-lang.com/documentation/file.SASS_REFERENCE.html#control_directives__expressions


joshuaogle:
👍 

Wonder why the test didn't catch this

tysongach:
@joshuaogle So…I think it works fine. My theory is that Sass has an official syntax, but accepts both because it’s probably a common error. I want to align to how they document its use, in case the non-official version is ever deprecated or goes unsupported.

PR #904, opened 2016-06-01
Run SassDoc on CircleCI

tysongach:
We use [SassDoc](http://sassdoc.com/) for our documentation. This PR introduces a step within CircleCI to run SassDoc and parse the documentation; if it comes across an error, the build will fail. It would look like this:

screen shot 2016-06-01 at 10 08 57

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,7 @@
+dependencies:
+  override:
+    - bundle install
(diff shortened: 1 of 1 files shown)


dgalarza:
LGTM

tysongach:
Thanks, @dgalarza!

PR #905, opened 2016-06-03
Add input:not([type]) selector to CSS output examples

ArmadilloXX:
Updating online docs according to actual behavior of addon.


tysongach:
@ArmadilloXX Thanks for this! 👍 

Sorry about Circle…we don’t have proper tests set up for this and we’re soon changing it, so the fail is fine for now.

PR #924, opened 2016-07-24
Update size addon

matheusagcosta:
### What does this PR do?

Refactor size addon

### DIFF
--- a/app/assets/stylesheets/addons/_size.scss
+++ b/app/assets/stylesheets/addons/_size.scss
@@ -29,23 +29,11 @@
 ///
 /// @todo Refactor in 5.0.0 to use a comma-separated argument
 
-@mixin size($value) {
-  $width: nth($value, 1);
-  $height: $width;
+@mixin size($s-width, $s-height: null) {
+  height: $s-width;
+  width: $s-width;
 
-  @if length($value) > 1 {
-    $height: nth($value, 2);
-  }
-
-  @if is-size($height) {
-    height: $height;
-  } @else {
-    @warn "`#{$height}` is not a valid length for the `$height` parameter in the `size` mixin.";
-  }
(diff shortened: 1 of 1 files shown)


tysongach:
Hi @matheusagcosta! Thanks for the PR. Couple notes:

1. What’s your goal with refactoring the `size` mixin? How is it better with these changes?
2. You changed the way the arguments work and set the default for `height` as `null`, instead of it equaling the width argument, as it was before. This removes the ability to pass one argument in, and get both width and height back (easily creates squares), like so:
  ```scss
  @include size(2rem);

  // output
  height: 2rem;
  width: 2rem;
  ```
The above would not be possible with your refactor, since `$s-height` is defaulted to `null`.
3. I also noticed you removed our “validiation” steps, which were those `@if` blocks that run the user’s values through our [`is-size` function](https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/validators/_is-size.scss). This function checks to make sure the user input a valid CSS length or keyword like `fit-content`, and if not, it throws a warning. These are important to help user’s find accidental mistakes. So if someone did this:

  ```scss
  @include size(foo);
  ```

  They would receive this warning:

  ```
  Error: `foo` is not a valid length for the `$height` argument in the `size` mixin. on line…
  ```

4. Finally, I wanted to mention that the `size` mixin has already been refactored in v5 (currently in beta) to add a comma between the two arguments. You can find that v5 version [on the `master` branch](https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/library/_size.scss). You’re refactor is against the v4 version, and we wouldn't be able to update that on the v4 release cycle since you have added a comma and that would be a breaking change.

tysongach:
Closing; lack of response.

PR #930, opened 2016-08-25
prepend inputs with parent selector through loop

opihana:
### What problem am I trying to solve?

I wanted $all-text-inputs to have styling when those inputs are nested within a certain class. This can currently be done by nesting: 

```
.parent-class {
  $all-text-inputs {
    ...
  }
}
```

However, it was less ideal when trying to expand those same styles to non-nested classes, while trying to maintain DRY Sass files. For example, we were looking to do:

```
.parent-class $all-text-inputs,
.cousin-class,
.maybe-something-else {
  ...
}
```

Clearly, this produces undesired css: the first text input would be prepended with the `.parent-class`, but none of the following inputs would. 

### What does this PR do?

This PR extends the existing functionality of the `_assign_inputs` function to accept a parent-selector that then gets prepended during the loop, in the same way the `pseudo` classes do.

### Critiques of this approach?

* Theoretically, we could go on forever, making this apply to child-selectors and so on. That could become onerous.
* One way around this is to add another class to the items identified in my example above. This is not ideal when using form auto-generation, like [Simple Form](https://github.com/plataformatec/simple_form).
* Another way around this is to add a local function that would loop through the inputs and prepend them. This is not ideal as the `$all-text-inputs` already has the comma separation. 
* Anything else?

### DIFF
--- a/core/bourbon/utilities/_assign-inputs.scss
+++ b/core/bourbon/utilities/_assign-inputs.scss
@@ -14,7 +14,8 @@
 
 @function _assign-inputs(
     $inputs,
-    $pseudo: null
+    $pseudo: null,
+    $parent-selector: null
   ) {
 
   $list: ();
@@ -22,6 +23,7 @@
   @each $input in $inputs {
     $input: unquote($input);
     $input: if($pseudo, $input + ":" + $pseudo, $input);
+    $input: if($parent-selector, $parent-selector + " " + $input, $input);
     $list: append($list, $input, comma);
   }
 
--- a/spec/bourbon/utilities/assign_inputs_spec.rb
+++ b/spec/bourbon/utilities/assign_inputs_spec.rb
@@ -18,7 +18,7 @@ describe "assign-inputs" do
     end
   end
 
-  context "expands text inputs with pseudo classes" do
+  context "pseudo class provided" do
     it "finds selectors" do
       list = @text_inputs_list.dup
       list.map! { |input| input + ":active" }
@@ -44,6 +44,28 @@ describe "assign-inputs" do
       list.unshift "[type=\"file\"]"
       list.each do |input|
         expect(input).to have_rule("color: #f0f")
+      end
+    end
+  end
+
+  context "parent selector provided" do
+    it "prepends parent selector to each input" do
+      list = @text_inputs_list.dup
+      inputs_with_parent_selector = list.map! { |input| ".class " + input }
+      inputs_with_parent_selector.each do |selector|
+        expect(selector).to have_rule("color: #ddd")
+      end
+    end
+  end
+
+  context "parent selector and psuedo class provided" do
+    it "prepends parent selector and appends pseudo class to each input" do
+      list = @text_inputs_list.dup
+      inputs_with_psuedo_class = list.map! { |input| input + ":active" }
+      inputs_with_parent_selector =
+        inputs_with_psuedo_class.map! { |input| ".class " + input }
+      inputs_with_parent_selector.each do |selector|
+        expect(selector).to have_rule("color: #eee")
       end
     end
   end
--- a/spec/fixtures/utilities/assign-inputs.scss
+++ b/spec/fixtures/utilities/assign-inputs.scss
@@ -17,3 +17,11 @@ select {
 #{_assign-inputs($_text-inputs-list)} {
   color: #f0f;
 }
+
+#{_assign-inputs($_text-inputs-list, null, ".class")} {
+  color: #ddd;
+}
+
+#{_assign-inputs($_text-inputs-list, active, ".class")} {
+  color: #eee;
+}


tysongach:
Hi @opihana 👋 !

Just to be clear, the CSS output you are trying to achieve is:

```css
.parent-class input[type="color"],
.parent-class input[type="date"],
.parent-class input[type="datetime"],
.cousin-class,
.maybe-something-else {
  ...
}
```

_(simplified for the sake of example)_

I worry that we’re building too specific of selectors and I suspect we’re doing that so to fight specificity issues that are introduced further up the cascade. Meaning, nesting base HTML type selectors within a class is usually a smell that our object-oriented CSS went wrong somewhere.

What is the specific reason we need to having the `input` selectors within the class in the first place? Does applying your styles just to the class not work?

```css
.parent-class,
.cousin-class,
.maybe-something-else {
  ...
}
```

> Theoretically, we could go on forever, making this apply to child-selectors and so on. That could become onerous.

I’m worried about that, too. I think this would also be the first time a Bourbon mixin would be aware of a user/project-based class, which I feel is anti-goal, or at least an indicator that we are compensating for an architectural problem.

opihana:
Hi! 👋 

> What is the specific reason we need to having the input selectors within the class in the first place? Does applying your styles just to the class not work?

It's actually an error class that is applied by our Simple Form setup. It wraps the input field, which includes more than the input itself, meaning: 

> Does applying your styles just to the class not work?

Nope. Here are screenshots and more specific code:

screen shot 2016-08-25 at 2 54 07 pm

```
.field-with-errors {
  #{all-text-inputs} {
    border-color: $error-color;
    yadda...
  }
}
```

screen shot 2016-08-25 at 3 01 12 pm


```
inline-field-group.group-with-errors {
  border-color: $error-color;
  yadda...
}
```

> nesting base HTML type selectors within a class is usually a smell that our object-oriented CSS went wrong somewhere.

Yup! Another option that I hadn't thought of before is updating our Simple Form config to add a class to every text input. Not all of our forms can use Simple Form, so I doubt we will go with this approach. 

> ...code smell...

It appears that we could extract a class of just `input-error` (or something) that has those common styles. We would then need to update our Simple Form config to add both the `input-error` class AND the appropriate error class for the input type. Again, not all of our forms can use Simple Form, so this would require an audit of those forms and add the classes manually where needed. Totally doable, though, and maybe the best approach for _unscented code_. 

> I think this would also be the first time a Bourbon mixin would be aware of a user/project-based class, which I feel is anti-goal. 

If so, just close it, and we'll come up with a better solution locally!

Thanks for the quick review.

P.S. As you can see, I made some pattern library example changes since you last saw it. If you can come up with a funny birthday input example, let me know. It's the only one left.

tysongach:
Ahh, Simple Form…

I’d definitely encourage configuring Simple Form and trying to reduce complex CSS selectors. I used this config on a project recently:

```ruby
SimpleForm.setup do |config|
  config.wrappers :default,
                  class: "c-form-field",
                  error_class: "c-form-field--has-error" do |b|
    b.use :html5
    b.use :placeholder
    b.optional :maxlength
    b.optional :pattern
    b.optional :min_max
    b.optional :readonly
    b.use :label, class: "c-form-field__label"
    b.use :input, class: "c-form-field__input"
    b.use :hint, wrap_with: { class: "c-form-field__hint" }
    b.use :error, wrap_with: { class: "c-form-field__error" }
  end

  config.boolean_label_class =
    "c-form-field__label c-form-field__label--boolean"
  config.boolean_style = :nested
  config.browser_validations = false
  config.button_class = "button"
  config.default_wrapper = :default
  config.error_notification_class = "error_notification"
  config.error_notification_tag = :div
  config.item_wrapper_tag = nil
  config.label_text =
    lambda { |label, required, _explicit_label| "#{label} #{required}" }
end
```

This still doesn’t give you an error class on the input, though. As far as I know, Simple Form can’t do that?

But ultimately, I’m curious why this…

```scss
.field-with-errors {
  #{all-text-inputs} {
    border-color: $error-color;
    yadda...
  }
}
```

…can’t be this:

```diff
.field-with-errors {
-  #{all-text-inputs} {
+  .form-input {
    border-color: $error-color;
    yadda...
  }
}
```

The point here is that `$all-text-inputs` is base HTML elements, which are great when used for establishing a global baseline look and feel for all inputs. But down to a component level (a field, and specifically a field with an error), we should just style on classes. So, can you style your input on a class like `.form-input`, for example, instead of `$all-text-inputs`.

Finally, I’m not sure these changes will accomplish your needs, because `$all-text-inputs` is just a variable, so it does not accept arguments. And `_assign-inputs` is a private function, so I don’t recommend using that directly.

Love your pattern library dummy data! 😉

tysongach:
@opihana I’m going to close this, but you know I’m always happy to chat about CSS 😉

opihana:
@tysongach, thanks for sharing your recent Simple Form config.

Yes, as far as I'm aware, we'd need some javascript to add an error class to the input.

Yes, I can use the code sample you provided, if I just add `.form-input` to my setup which will be quite easy. 👍 

And yes, I would have to reassign the `$all-text-inputs` variable, which is absolute nonsense. 

Thanks again for the review and the thorough explanation of your thoughts!

PR #935, opened 2016-09-05
improve contrast-switch

xi:
I tried to improve the contrast-switch function:
- `$light-color` and `$dark-color` are taken into account when determining whether `$base-color` is light.
- The algorithm for calculating the contrast conforms with [WCAG](https://www.w3.org/TR/WCAG20/#contrast-ratiodef) (though it is not exact because Sass does not have a `pow()` function)

See also less/less.js#2754 for a similar change in less.

I did not yet include any new tests because I do not have a working ruby setup right now. Please let me know what you think of this. If you like it I will manage to add some tests somehow.


tysongach:
Hey @xi! Thanks for the PR!

I’m super excited to get this as near as conforming to WCAG as we can. I haven’t had much time to fully review yet, but I did notice that even though this is to “improve contrast-switch”, the actual `contrast-switch` function is untouched in this PR. Is this intentional?

core/bourbon/helpers/_lightness.scss
```diff
@@ -0,0 +1,36 @@
+@charset "UTF-8";
+
+@function _gamma($channel) {
```
tysongach: We should move this function into its own partial and also add SassDoc documentation.

core/bourbon/helpers/_contrast.scss
```diff
@@ -0,0 +1,31 @@
+@charset "UTF-8";
+
+/// Programatically determines the contrast between two colors.
+///
+/// Note that the alpha channel is ignored.
+///
+/// @link https://www.w3.org/TR/WCAG20/#contrast-ratiodef
```
tysongach: We’ve been consistently using short URL’s for `@link`, so this can be `https://goo.gl/54htLV`.

core/bourbon/helpers/_contrast.scss
```diff
@@ -0,0 +1,31 @@
...
+/// @return {number (1-21)}
+///
+/// @example scss
+///   contrast(black, white)
+///
+/// @require {function} _lightness
+///
+/// @access private
+
+@function _contrast($color-1, $color-2) {
```
tysongach: I think naming this function `_contrast-ratio` provides a bit more clarity as to what it outputs.

core/bourbon/helpers/_lightness.scss
```diff
@@ -0,0 +1,36 @@
...
+  @if $channel < 0.03928 {
+    @return $channel / 12.92;
+  } @else {
+    $c: ($channel + 0.055) / 1.055;
+    // very similar to pow($c, 2.4)
```
tysongach: This comment could become part of SassDoc documentation comments, if useful. We tend to avoid inline Sass comments.

tysongach:
We should also add a note to the documentation for `_contrast-switch` that it conforms to WCAG.

xi:
Hi @tysongach, thanks for the review! I implemented most of the changes you requested.

> We should also add a note to the documentation for `_contrast-switch` that it conforms to WCAG.

I am not sure if this is a good idea. While it uses the contrast definition of WCAG, it does not at all guarantee a minimum contrast ratio. So "conforming" may be too big a word here.

tysongach:
> I am not sure if this is a good idea. While it uses the contrast definition of WCAG, it does not at all guarantee a minimum contrast ratio. So "conforming" may be too big a word here.

Great point. Conforming is a strong word, so maybe we say it _aims_ to be WCAG 2.0 AA compliant, but that the function does not map directly to how that is defined. I’ll add this to the docs.

core/bourbon/helpers/_contrast-ratio.scss
```diff
@@ -0,0 +1,31 @@
...
+/// @link https://goo.gl/54htLV
+///
+/// @argument {color (hex)} $color-1
+///
+/// @argument {color (hex)} $color-2
+///
+/// @return {number (1-21)}
+///
+/// @example scss
+///   contrast(black, white)
```
tysongach: Needs updating to new name: 

```diff
- ///   contrast(black, white)
+ ///   _contrast-ratio(black, white)
```

core/bourbon/helpers/_gamma.scss
```diff
@@ -0,0 +1,21 @@
+@charset "UTF-8";
+
+/// Perform gamma correction on a single color channel.
+///
+/// Note that sass does not have a pow() function, so the
```
tysongach: `sass` → `Sass`

tysongach: Wrap `pow()` in back ticks to represent code.

core/bourbon/helpers/_lightness.scss
```diff
@@ -0,0 +1,24 @@
+@charset "UTF-8";
+
+/// Programatically determines the lightness of a color.
+///
+/// @argument {color (hex)} $hex-color
+///
+/// @return {number (0-1)}
+///
+/// @example scss
+///   lightness($color)
```
tysongach:
```diff
- lightness($color)
+ _lightness($color)
```

core/bourbon/helpers/_gamma.scss
```diff
@@ -0,0 +1,21 @@
+@charset "UTF-8";
+
+/// Perform gamma correction on a single color channel.
```
tysongach:
```diff
- Perform
+ Performs
```

tysongach: Maybe we should explain from a high-level what gamma correction is?

xi: Yes, I also thought about that. Unfortunately, I am still very confused about the exact terminology (see for example [this discussion](https://github.com/less/less.js/issues/2829)). So instead of getting the details wrong I just skipped them.

If you have a good idea just go ahead.

xi:
Oh, most of this was just my sloppiness. Sorry!

Note that https://github.com/thoughtbot/bourbon/pull/935#discussion_r85646986 is now collapsed but still relevant.

core/bourbon/helpers/_gamma.scss
tysongach: @xi Makes sense! We can loop back later if one of us figures it out 😉

tysongach:
> Oh, most of this was just my sloppiness. Sorry!

@xi No worries, at all! Thank you _so_ much for your work here!

tysongach:
Squashed and merged as 4c0fdd810415b57e4d8ebdc08e63b266179abdab.

@xi Thank you for your work on this! Super happy about this refinement. 🎉

xi:
Thanks for merging!

Issue #937, opened 2016-09-11
Broken image in the README


sobolevn:
### Screenshots (if relevant)
<img width="868" alt="2016-09-11 15 21 30" src="https://cloud.githubusercontent.com/assets/4660275/18417280/a811a558-7833-11e6-941d-b0b33b66f884.png">

PR #938, opened 2016-09-12
Fix logo on readme

whmii:
See: https://github.com/thoughtbot/bourbon/issues/937


tysongach:
Ah, yes we should fix this! Thanks, Will!

tysongach:
Merged as b7d1619afe42104d3116bd08f0b2a0eb76f9867f.

PR #947, opened 2016-10-25
Update position mixin

phlppschrr:
### What does this PR do?
Fixes position mixin when called with only a single length value. 

this works now, and failed before:
`.x { @include position(0); }`

### If this is related to an existing issue, include a link to it as well.
no related issue found

### DIFF
--- a/core/bourbon/library/_position.scss
+++ b/core/bourbon/library/_position.scss
@@ -30,7 +30,7 @@
     $coordinates: null
   ) {
 
-  @if type-of($position) == list {
+  @if type-of($position) == list or _is-length($position) {
     $coordinates: $position;
     $position: relative;
   }


tysongach:
Hi @phlppschrr! Thanks for the PR.

I’m not sure I fully understand the change yet and have a few questions:

1. What is the expected output of your example?

  ```scss
  @include position(0);
  ```

1. Our [`position` mixin](http://bourbon.io/docs/#position) is currently designed to take two arguments. The first is to define the value for the `position` property, the second accepts values for `top`, `right`, `bottom` and `left` properties, and it can accept shorthand and `null` values to fill those out.

  ```scss
  @include position(relative, 0 null null 10em);

  // output
  position: relative;
  top: 0;
  left: 10em;
  ```

  If the mixin only takes one argument that is a number, how do we define a `position` value?

1. You’re checking for a [CSS length](https://developer.mozilla.org/en-US/docs/Web/CSS/length) (e.g. `10px` or `50vw`) via our `_is-length` function, but what is being passed in is a `position` value (`static`, `relative`, etc.). Is this as intended?

phlppschrr:
Thanks for reviewing my PR.

You are right, it seems my patch is fixing an undocumented behaviour, but my changes are indeed intended.

Currently if you pass the $coordinates as the first parameter the mixin is using`$position: relative' as a fallback and is assuming the first parameter is a list of coordinates. A single length value works if passed as the second parameter.

1. Expected output of `@include position(0)` would be
```
  position: relative;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
``` 
2. `position: relative`is already the fallback default value
3. This is intended. It is just fixing current undocumented (but useful) behaviour.

I created a sassmeister example to illustrate: http://www.sassmeister.com/gist/edef15d880468c23650c05b54831eee7

tysongach:
Ah, gotcha! This makes sense.

So we were already defining a default `position` value in the mixin arguments, but you can’t currently _use_ that default. Your change here fixes that.

I’m actually mixed on this. While this is a great catch and a working change, I wonder if using the default _masks)_ what the CSS will be output. We like for Bourbon’s features to be obvious.

This Sass:

```scss
@include position(0);
```

…is a _bit_ unclear as to what is actually output, compared to this:

```scss
@include position(relative, 0);
```

So while it is a little more verbose, I think it provides clarity. Maybe we actually remove the default…

One other thing you mention that we should do is check the `$coordinates` argument list for whether they are CSS lengths or not, and provide helpful warnings.

whmii:
Agree, If I had to guess I would think more people would think that `absolute` would be the default. I think, preferences aside, that having it be explicit makes it more "future proof" and avoids other PRs like "`position` mixin defaults to relative but I think X would be better, lets change it."

tysongach:
Here’s a PR which removes the default value of `relative` for `$position`. I also removed the default value for `$coordinates`, because just putting `@include position(relative);` is also unclear that it outputs `0`’s.

phlppschrr:
Closing my PR, as https://github.com/thoughtbot/bourbon/pull/948 makes more sense for Bourbon.

tysongach:
Thanks for the PR and discussion, @phlppschrr!

PR #948, opened 2016-10-26
Remove argument defaults from position mixin

tysongach:
- Remove default value for the `$position` argument because it is
unclear if it is not defined when calling the mixin.
- Remove default value for the `$coordinates` argument because it is
unclear if it is not defined when calling the mixin.
- This also updates the documentation to show use of shorthand.
- Remove unused spec CSS.
- Related discussion: https://github.com/thoughtbot/bourbon/pull/947.

--

TL;DR

The mixin is currently setup to do this:

```scss
@include position(relative);

// output
position: relative;
top: 0;
right: 0;
bottom: 0;
left: 0;
```

…and:

```scss
@include position(0 10px);

// output
position: relative;
top: 0;
right: 10px;
bottom: 0;
left: 10px;
```

But these are unclear by not declaring _both_ arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ project adheres to [Semantic Versioning](http://semver.org).
 
 ## [Unreleased]
 
+### Changed
+
+- Removed the default values from the `$position` and `$coordinates` arguments
+  for the `position` mixin.
+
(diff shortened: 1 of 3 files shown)


kylefiedler:
+1 love that this adds more clarity

Issue #944, opened 2016-10-14
Update `hide-visually` for better pronunciation


tysongach:
Facebook recently discovered that the technique they were using to hide content visually (which is the same as what we use in the `hide-visually` mixin) was causing each letter to wrap to new lines and therefore be pronounced oddly.

Adding `white-space: nowrap;` resolved the issue, and we should update our mixin to do the same.

Reference: https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe#.z5l8ex9ed

bekapod:
@tysongach were you going to work on this or can I work on a PR (within the next few days)? 😄

tysongach:
@bekapod Go for it! Thanks for hopping in.

bekapod:
(I haven't forgotten about this, I plan to work on it this evening 😄 )

PR #949, opened 2016-10-26
Improve accessibility on hide-visually mixin

bekapod:
Add `white-space: nowrap;` to hide-visually mixin, and reset it back to `inherit` when unhiding.

Screen readers don't interpret line breaks as spaces, so force visually hidden content onto one line so spaces are retained and screen readers can read the content properly.

See #944


tysongach:
👍  Thanks, @bekapod!

tysongach:
Thanks again, @bekapod! Merged as 2fc2f36e5868688baec1641e2b5d9afe044d2a62.

PR #952, opened 2016-10-26
Update `$font-stack-system` list

tysongach:
### What does this PR do?
- Updates the `$font-stack-system` list to include more fonts.
- Follows Adam Morse's system-fonts: https://github.com/mrmrs/css-system-fonts
### To do
- [x] Update Change Log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,12 @@ project adheres to [Semantic Versioning](http://semver.org).
   full compliance, though all of our manual testing passed.
 - Renamed the `$coordinates` argument in the `position` mixin
   to `$box-edge-values`.
+- Updated `$font-stack-system` to include Avenir Next, Avenir, Lucida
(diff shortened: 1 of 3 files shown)


kylefiedler:
I'm not sure I agree with Avenir going in this list. It's not really a UI typeface (Neither is Helvetica but it's at least used that way on older iOS and MacOS versions).

I guess what I'd want is to have a font stack that mirrored their system OS font and it doesn't.

tysongach:
@kylefiedler In the context of this list, I see Avenir the same as Helvetica. iOS started shipping with it a few years back and I had a nice experience [using it](https://dribbble.com/tysongach/projects/189658-Involvio-for-iOS).

I think it’s there in the case that San Francisco isn’t available, but if Avenir is, use that instead of Helvetica. I also think Avenir is closer in style to SF than Helvetica, so it plays the middle-ground.

PR #956, opened 2016-10-28
Clarify `directional-property` argument names and docs

tysongach:


whmii:
always a fan of removing abbreviations when they are unnessisary. LGTM

PR #968, opened 2016-11-03
Fixing "user-select" mixin on Chrome for Android

benzaita:
Fixes the `user-select` mixin in case `value==none` and running on Chrome on a touch device.

### DIFF
--- a/app/assets/stylesheets/css3/_user-select.scss
+++ b/app/assets/stylesheets/css3/_user-select.scss
@@ -2,4 +2,9 @@
   @include _bourbon-deprecate-for-prefixing("user-select");
 
   @include prefixer(user-select, $value, webkit moz ms spec);
+
+  // See http://stackoverflow.com/a/30319808/21290
+  @if $value == none {
+    -webkit-tap-highlight-color: transparent;
+  }
 }


tysongach:
Hi @benzaita! Thanks for the PR.

I don’t think this is a good change for Bourbon for a couple of reasons:

1. The `-webkit-tap-highlight-color` property is [non-standardized](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color).
2. As seen on Line 2 in the partial, the `user-select` mixin is deprecated (in the soon-to-be-released v4.3) and completely removed in v5.0 (currently in beta). This is because Bourbon will no longer provide mixins that only handle prefixing. Instead, we are encouraging use of Autoprefixer or other automated tools.
3. I’m also hesitant to hard code the `transparent` value as Bourbon tries to carry as little visual opinion as possible. I imagine some would want to be able to set that color dynamically and then we end up conflating this small mixin which is really just for prefixing.
4. I’m unclear if this is related for sure, but it’s worth noting that Can I Use says that `user-select` is “Reported to not work in some manufacturer's versions of Android 4.0 and below, though confirmed to work in others.” Is it possible this is just an Android bug?

tysongach:
No response; closing.

PR #978, opened 2016-11-15
Add style property to package.json, allowing for easier imports.

davejtoews:
When installed via npm, a style property in package.json allows this module to be easily imported when using tools such as [npm-sass](https://github.com/lennym/npm-sass) or [sass-module-importer](https://www.npmjs.com/package/sass-module-importer).

Apparently the "style" key is not a standard, but it is a growing trend. Here's a StackOverflow thread on the subject: http://stackoverflow.com/questions/32037150/style-field-in-package-json

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   ],
   "license": "MIT",
   "main": "index.js",
+  "style": "app/assets/stylesheets/_bourbon.scss",
   "name": "bourbon",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   ],
   "license": "MIT",
   "main": "index.js",
-  "style": "app/assets/stylesheets/_bourbon.scss",
+  "style": "core/_bourbon.scss",
   "name": "bourbon",
   "repository": {
     "type": "git",


tysongach:
I’m hesitant to add non-standard fields…

Have these changes been tested with npm-sass and sass-module-importer?

davejtoews:
@tysongach Yes, I've tested with both. The same can be said for the similar PR I made on Neat.

davejtoews:
@tysongach I get your hesitation to use a non-standard field, but I'm not really seeing the downside. It shouldn't affect any projects that don't make use of it.

[Bootstrap added "style" to package.json two and a half years ago.](https://github.com/twbs/bootstrap/pull/12441) They chose to point to their compiled css file rather than a pre-processor source in order to support the majority use case. In the case of Bourbon, all your users are Sass users of course, so adding `"style": "core/_bourbon.scss"` should have no effect except to allow for easier use among a subset of your users.

tysongach:
> but I'm not really seeing the downside. 

I’m mostly concerned with maintenance. I know it’s just “one line”, but there are many of these “one lines” sprinkled throughout for other third-party integrations that we have to maintain when things change and break. Since it is not standardized, we can’t follow a systemized approach.

The fact that npm-sass and sass-module-importer both use `style` is likely a happy coincidence. If we did add a field, I’m more inclined to add a `sass` field because that is what Bourbon is: a collection of Sass partials. We don’t have any compiled CSS. But it seems that that would leave sass-module-importer out of the picture.

I totally see where you’re coming from, but it’s tough as a library to support “all the things” and strive to be bug-free.

davejtoews:
>The fact that npm-sass and sass-module-importer both use `style` is likely a happy coincidence.

More than a coincidence, I think. [npm-css](https://github.com/defunctzombie/npm-css), [rework-npm](https://github.com/reworkcss/rework-npm), [npm-less](https://github.com/Raynos/npm-less), [parcelify](https://github.com/rotundasoftware/parcelify), and  as @Hackzzila suggested above [diamond](https://github.com/diamondpkg/diamond) all use it too. Of course only the latter two apply to Sass.

Though not an official standard, there seems to be a lot of support for 'style'.

>I’m more inclined to add a sass field because that is what Bourbon is: a collection of Sass partials. ...  But it seems that that would leave sass-module-importer out of the picture.

I'll note that both npm-sass and sass-module importer support using a sass file for 'main' as well as for 'style'. I'm not sure how this would affect using Bourbon with other tools, but since as you point out Bourbon is a collection of _Sass_ partials, there's something a bit illogical about having a _javascript_ file as it's `main`.

tysongach:
> there's something a bit illogical about having a javascript file as it's main.

Definitely, and it is something I’ve never really been happy with.

We used to set `main` to point to the primary Bourbon manifest Sass file, but people still had to manually set that path up. It might be time to revisit all of that, as a whole.

Hackzzila:
I am in support of either a `sass` or `style` field. `sass` seems more logical, but `style` has more support. Overall I don't really care much, but I am leaning towards `sass.`

davejtoews:
One last point in support of the `style` field. [It's been flagged as a "nice-to-have" feature in webpack's css-loader to make use of it.](https://github.com/webpack-contrib/css-loader/issues/51). If webpack picks it up (and adds it to sass-loader too) then i'd suggest it might as well be a standard.

Hackzzila:
What is the status on this?

tysongach:
Thanks, @davejtoews, merged as https://github.com/thoughtbot/bourbon/commit/c10f0bf18feafa866a51259b65c765642c62bf8c.

Issue #994, opened 2017-01-28
Using triangle mixin with one size value in 5.0


Trost123:
In version 4 I was creating two triangles, using:

```scss
@include triangle(9px, gray, down);
@include triangle(9px, white, right);
```

And it resulted in what you'd expect: right and down triangles of the same size.
http://prntscr.com/e1eo6z

In version 5 I changed the parameter order to fit new specs:

```scss
@include triangle(down, grey, 9px);
@include triangle(down, white, 9px);
```

http://prntscr.com/e1epai

The triangles with direction-right get squashed vertically now.

```scss
@include triangle(right, grey, 4px, 9px);
```

Manually imputing width and height fixes the triangle to look like in screenshot 1, but it looks like a downgrade from v4 behavior. If you input one value for a triangle - it should be it's biggest side (hypotenuse), no matter what orientation it is.
Same for diagonal ones: if I try to create down-right triangle, 9px size, I get way bigger triangle, than a down-oriented triangle 9px size.

tysongach:
Hi @Trost123!

The aim by re-working the mixin was to make the argument values being set more clear and expected.

> If you input one value for a triangle - it should be it's biggest side (hypotenuse), no matter what orientation it is.

This is nicety when the mixin was called, however, I think it is an assumption that everyone would see the one value as always being the longest side.

With the new mixin, the width and height set will always map directly to the width and height of the "box" the triangle sits in. I like the clarity of that, because that is the actual space it takes up when rendered. In the old mixin, if you pass one value, you’d have to know that that value is either the width _or_ the height, depending on the value set for the direction. 

I’m now wondering if we should remove the default values from the arguments so that authors have to define values for the four arguments each time the mixin is called.

Trost123:
I think 90% uses of triangle mixin are for [isosceles right triangle](http://mathworld.wolfram.com/IsoscelesRightTriangle.html) and not that [squashed grey thing](http://prnt.sc/e1epai) I got in a screenshot.
So, while I agree with "box" size concept you have in mind, why is box height = width/2 for left\right triangles right now?

I always imagined a html triangle as a 1\4 segment of a square pizza](http://prntscr.com/e1lqas).
And the images in old documentation worked for me as soon as I saw it.
That's why I never knew about the fourth parameter when using Bourbon v4.
Size = "bounding square" (not rectangle) side.

Here are variations I came up with for:
`triangle(10px, color, side);`
1) Bourbon 4 style:
[...]

PR #998, opened 2017-02-05
Remove triangle mixin’s default argument values

tysongach:
- These values are very use-case specific, e.g. matching the color of
whatever component it is used on.
- The size arguments also acted unexpectedly because the shape of the
triangle when defining only the width would be different based on if it
was the up/down or right/left direction.
- By requiring all four arguments to be defined, it forces consistency
and clarity.
- The order of the arguments was also adjusted so that those relating to
the "box" the triangle sits in are grouped.
- Closes https://github.com/thoughtbot/bourbon/issues/994

## Before

```scss
@include triangle("up", #b25c9c, 2rem);
```

## After

```scss
@include triangle("up", 2rem, 1rem, #b25c9c);
```

### DIFF
--- a/core/bourbon/library/_triangle.scss
+++ b/core/bourbon/library/_triangle.scss
@@ -2,23 +2,23 @@
 
 /// Generates a triangle pointing in a specified direction.
 ///
-/// @argument {string} $direction [up]
+/// @argument {string} $direction
 ///   The direction the triangle should point. Accepts `up`, `up-right`,
 ///   `right`, `down-right`, `down`, `down-left`, `left` or `up-left`.
 ///
-/// @argument {color} $color [currentColor]
-///   Color of the triangle.
-///
-/// @argument {number (with unit)} $width [1rem]
+/// @argument {number (with unit)} $width
 ///   Width of the triangle.
 ///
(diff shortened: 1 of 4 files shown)


core/bourbon/library/_triangle.scss
```diff
@@ -2,23 +2,23 @@
 
 /// Generates a triangle pointing in a specified direction.
 ///
-/// @argument {string} $direction [up]
+/// @argument {string} $direction
```
whmii: woof

spec/fixtures/library/triangle.scss
```diff
@@ -1,9 +1,9 @@
 @import "setup";
 
-.triangle {
-  @include triangle;
+.triangle--up {
+  @include triangle("up", 2rem, 1rem, #b25c9c);
```
whmii: you don't need these quotes here (i think). doesn't change anything tho

tysongach: That’s correct. I prefer putting strings within quotes.

whmii:
I'm down with this. it's odd to have defaults for such a particular implementation.

PR #1016, opened 2017-03-21
Function strip-units - update warn message

### LINKED ISSUE #1013: [Deprecation] Function strip-unit is not available in v4.3.3
### Expected behavior
Include `strip-unit` function in Bourbon v4.3 because function `strip-units` function  is deprecated since v4.3.3 .

But I cannot found definition of `strip-unit` function…
### Steps to reproduce behavior
```sh
git clone https://github.com/thoughtbot/bourbon.git -b v4.3.3
cd bourbon
git grep '@function strip-units\b' -- *.scss # 1 result
git grep '@function strip-unit\b'  -- *.scss # no result!!!
```
### Bourbon version
v4.3.3

stonio:
Update warn message about `strip-units()` function to close #1013.

### DIFF
--- a/app/assets/stylesheets/functions/_strip-units.scss
+++ b/app/assets/stylesheets/functions/_strip-units.scss
@@ -15,7 +15,7 @@
 @function strip-units($value) {
   @if $output-bourbon-deprecation-warnings == true {
     @warn "[Bourbon] [Deprecation] `strip-units` is deprecated and will be " +
- "removed in version 5 and replaced by `strip-unit` function.";
+    "removed in 5.0.0 and replaced by the `strip-unit` function.";
   }
 
   @return ($value / ($value * 0 + 1));
--- a/app/assets/stylesheets/functions/_strip-units.scss
+++ b/app/assets/stylesheets/functions/_strip-units.scss
@@ -14,8 +14,8 @@
 
 @function strip-units($value) {
   @if $output-bourbon-deprecation-warnings == true {
-    @warn "[Bourbon] [Deprecation] `strip-units` is deprecated and will be " +
-    "removed in 5.0.0. Use the renamed `strip-unit` function instead.";
+    @warn "[Bourbon] The `strip-units()` function will be " +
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/functions/_strip-units.scss
```diff
@@ -14,8 +14,8 @@
 
 @function strip-units($value) {
   @if $output-bourbon-deprecation-warnings == true {
-    @warn "[Bourbon] [Deprecation] `strip-units` is deprecated and will be " +
```
tysongach: This function is deprecated, so I think we should leave in `[Deprecation]` and `is deprecated`.

According to [Wikipedia](https://en.wikipedia.org/wiki/Deprecation#Software_deprecation):

> …deprecated status may also indicate the feature will be removed in the future.

As you mentioned in #1013, “Use the renamed `strip-unit` function instead” is a bit confusing because you can’t use that _yet_ if you are still on 4.x. So we should re-word only that sentence to make it clear that you will need to use the renamed function when you upgrade to 5.0.

app/assets/stylesheets/functions/_strip-units.scss
```diff
@@ -15,7 +15,7 @@
 @function strip-units($value) {
   @if $output-bourbon-deprecation-warnings == true {
     @warn "[Bourbon] [Deprecation] `strip-units` is deprecated and will be " +
-    "removed in 5.0.0. Use the renamed `strip-unit` function instead.";
+ "removed in version 5 and replaced by `strip-unit` function.";
```
tysongach: Can we indent this line as it was before (to line up with `@warn`?

And swap `version 5` for `5.0.0`, so that we speak in specific version numbers?

Finally, I think we are missing the word `the` before `strip-unit`.

tysongach:
Thanks, @stonio! Merged as https://github.com/thoughtbot/bourbon/commit/04084194ec2c40f9c2e5a397d699b0bc3006ecf6.

PR #1038, opened 2017-09-26
improve gamma approximation

xi:
## Description

In #935 I added a WCAG compliant contrast implementation. In order to calculate gamma corrected values I used an approximation of `pow($c, 2.4)`. Unfortunately, I later realized that this approximation is not as good as I had thought.

The approxomation is actually optimal if you only look at `_gamma()`. But the resulting values are later used to calculate a contrast, which amplifies certain errors made in this step. So I changed two things about the approximation:

- use `pow()` if available (e.g. from sass-planifolia or MathSass)
- use an approximation that is optimized to produce good contrast values

## Additional Information

See also https://github.com/oddbird/accoutrement-color/pull/6 for a similar change.

### DIFF
--- a/core/bourbon/utilities/_gamma.scss
+++ b/core/bourbon/utilities/_gamma.scss
@@ -2,8 +2,7 @@
 
 /// Performs gamma correction on a single color channel.
 ///
-/// Note that Sass does not have a `pow()` function, so the calculation
-/// is approximate.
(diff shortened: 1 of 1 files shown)


tysongach:
Thanks @xi! Merged as https://github.com/thoughtbot/bourbon/commit/5ac2d26329974ea928e72d8f816a6e53423f9807.

PR #1040, opened 2017-10-13
Update docs for changes to size add-on

killion:
## Description

The code now uses comma separated params but the docs showed it without them. This updates the docs.


tysongach:
Hi @killion!

Thanks for the PR. This documentation is correct as-is. The comma was introduced in v5 (currently in beta), but these are the docs for v4.

We have our WIP docs for v5 at http://bourbon.netlify.com/docs/latest/, which has the comma. Once we ship the stable release, these docs will consume the ones currently living at http://bourbon.io/docs/.

PR #1043, opened 2017-11-13
step numbers should be more than step one :)

captDaylight:


tysongach:
Hi! Thanks for the PR!

These get rendered by the Markdown parser as a proper numerically ordered list. By using `1` over an over, if you ever add or remove a step, you don’t have to re-number the whole list manually. It’s a small maintenance win. I’m going to leave as-is and close this PR.

PR #1069, opened 2019-01-16
Remove sass dependency

### LINKED ISSUE #1064: Remove (Ruby) sass dependency and switch to sassc for Bourbon 6.0.0?
Hello. 👋 

I was upgrading my dependencies on my project today and noticed that I'm still installing `sass` gem despite I already have `sassc`. After further inspection, it seems like that Bourbon specifies `sass` as a runtime dependency.

Given that [Ruby sass is now deprecated](http://sass.logdown.com/posts/7081811) and they suggested every one to move to `sassc`, I think it's best for Bourbon to stop specifying `sass` as a runtime dependency. 

I think there're two ways we could approach this issue:

* Switch to specify `sassc` as a runtime dependency.
* Remove `sass` from a runtime dependency, and raise an error if neither `sass` nor `sassc` is found.

Please let me know what you preferred. I'd be happy to submit a patch as well.

Thank you.

sikachu:
This closes #1064 

Given that Ruby sass is now deprecated and developers are suggested to move to sassc, we should stop specifying sass as a runtime dependency.

Also, since it seems like we required sass only just to add to load_paths, it should be safe to silently fail if it is not available.

Note that there's some caveat about this patch: sassc (or, precisely, sassc-ruby) no longer has `load_paths` setting at the class level. User has to set load_paths themselves [when they create `SassC::Engine` instance](https://github.com/sass/sassc-ruby/blob/61a02d1d88b489f83b5b3d0da48a9b62f210691b/test/custom_importer_test.rb#L81). This probably will affect people who are using sass in ruby but without Rails environment. I currently see no mention of this specific usage the README.md, so I think we shold be fine. (People who uses Rails and Asset Pipeline or Webpack should see no problem as they Railtie continues to work as expect).

### DIFF
--- a/bourbon.gemspec
+++ b/bourbon.gemspec
@@ -8,7 +8,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rake", "~> 11.1"
   s.add_development_dependency "rspec", "~> 3.4"
   s.add_development_dependency "scss_lint", "0.48"
-  s.add_runtime_dependency "sass", "~> 3.4"
   s.add_runtime_dependency "thor", "~> 0.19"
   s.authors = [
     "Christian Reuter",
--- a/lib/bourbon.rb
+++ b/lib/bourbon.rb
@@ -1,4 +1,3 @@
-require "sass"
 require "bourbon/generator"
 
 module Bourbon
@@ -7,6 +6,9 @@ module Bourbon
       config.assets.paths << File.expand_path("../../core", __FILE__)
     end
   else
-    Sass.load_paths << File.expand_path("../../core", __FILE__)
+    begin
+      require "sass"
+      Sass.load_paths << File.expand_path("../../core", __FILE__)
+    rescue LoadError; end
   end
 end


aadityataparia:
@houndci-bot Back again?

sikachu:
@tysongach would you mind giving this a look when you have a chance?

By the way, I always have a problem with Hound since after I left thoughtbot that I'll never get a check status back, not really sure why. Maybe you can run the lint from your side.

luccasmaso:
Good! Waiting for it

tysongach:
Quick update here: We spent a bit of time testing this out today, and while it’s fine in Rails, we are indeed hitting the caveat mentioned above. Outside Rails—for example in a Middleman app—these changes create a Sass import error, as Bourbon cannot be found.

Middleman 4.3 has moved off Ruby Sass and uses SassC, but without it supporting `load_paths`, I'm unsure how to set Bourbon up for those situations. I'm hesitant to merge this until we have a better grasp on that front.

I was using our [`apprentice.thoughtbot.com` repo](https://github.com/thoughtbot/apprentice.thoughtbot.com) as a test ground for this, in case anyone wants to try it.

tysongach:
Thanks @sikachu! This has been merged as https://github.com/thoughtbot/bourbon/commit/9bedac008832cb779ded2f9298f12a5ebae9e88a.

PR #1072, opened 2019-02-26
Swap scss-lint for stylelint

tysongach:
scss-lint has been deprecated.


tysongach:
The test failure is unrelated to these changes; this is fine to be merged without that getting resolved.

PR #1087, opened 2019-07-22
 Add `var()` and `env()` as valid CSS lengths

### LINKED ISSUE #1086: CSS custom properties skipped in position()
## Description

A CSS custom property used as edge value in the `position()` mixin is skipped.

## Steps to Reproduce

```scss
.my-selector {
  --my-var: 10rem;

  @include position(absolute, 5rem var(--my-var) null null);
}
```

Output:
```scss
.my-selector {
  --my-var: 10rem;

  position: absolute;
  top: 5rem;
}
```

Expected:
```scss
.my-selector {
  --my-var: 10rem;

  position: absolute;
  top: 5rem;
  right: var(--my-var); // this one is missing
}
```

Under the hood, it happens because the `position` mixin [ensures the edges are proper CSS lengths using `is-length`](https://github.com/thoughtbot/bourbon/blob/6ef70918ccdc0b5b049180a3071e4f978b9d9a99/core/bourbon/library/_position.scss#L58).

[`is-length` allows `calc()`](https://github.com/thoughtbot/bourbon/blob/04ac693bda30294ecb359aa97c43a3ee1b97c2a7/core/bourbon/validators/_is-length.scss#L12), but unfortunately it doesn’t allow [`var()`](https://developer.mozilla.org/en-US/docs/Web/CSS/var) nor [`env()`](https://developer.mozilla.org/en-US/docs/Web/CSS/env).

Would you accept a PR for this?

I’m not a fan of enforcing values (at least for `position`): if the value isn’t valid, the browser will naturally skip it. I think the responsibility for validation should be on the CSS author side, not on Bourbon side.

## Development Environment

Bourbon 5.0, MacOS, node-sass.

meduzen:
This PR closes #1086, and include tests, despite the fact I couldn’t accomplish the step 2 of the [PR contributing guidelines](https://github.com/thoughtbot/bourbon/blob/master/CONTRIBUTING.md#pull-requests) (I don’t know what `bundle` is, and it appears not to be on my computer).

Now, writing
```scss
@include position(absolute, 5rem var(--my-var) null null);
```
correctly output the `right: var(--my-var);` part instead of skipping it.

Waiting for feedbacks. Feel free to iterate upon this PR. Thanks for your work on Bourbon. 👍

### DIFF
--- a/core/bourbon/validators/_is-length.scss
+++ b/core/bourbon/validators/_is-length.scss
@@ -9,7 +9,12 @@
 /// @access private
 
 @function _is-length($value) {
-  @return type-of($value) != "null" and (str-slice($value + "", 1, 4) == "calc"
-       or index(auto inherit initial 0, $value)
-       or (type-of($value) == "number" and not(unitless($value))));
+  @return type-of($value) != "null"
+    and (
+      str-slice($value + "", 1, 4) == "calc"
+      or str-slice($value + "", 1, 3) == "var"
+      or str-slice($value + "", 1, 3) == "env"
+      or index(auto inherit initial 0, $value)
+      or (type-of($value) == "number" and not(unitless($value)))
+   );
 }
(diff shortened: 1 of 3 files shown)


tysongach:
Thanks so much, @meduzen! This has been merged as https://github.com/thoughtbot/bourbon/commit/eb23e19bd45aba7e777fbb45cc28eadff92067b3.

PR #1091, opened 2019-10-25
Update package.json

### LINKED ISSUE #1078: can't resolve`bourbon`
Some bundlers like `parcel` can't resolve`bourbon` when included in scss because of the `main` entry in `package.json` file. Please change the main entry point to same path as it is in `style` = `"core/_bourbon.scss",` :notes:

No result in scss looks like
```scss
@import 'normalize.css';
@import '~node_modules/bourbon/core/bourbon';
```
but could be like
```scss
@import 'normalize.css';
@import 'bourbon';
```

_Originally posted by @FDiskas in https://github.com/thoughtbot/bourbon/issues/802#issuecomment-510624302_
Reference: https://github.com/parcel-bundler/parcel/issues/3205

FDiskas:
Closes #1078

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "scss"
   ],
   "license": "MIT",
-  "main": "index.js",
+  "main": "core/_bourbon.scss",
   "style": "core/_bourbon.scss",
   "name": "bourbon",
   "repository": {


tysongach:
Can you please provide more detail and documentation around this change? I understand the issue to be around using Parcel with Bourbon; if you know of documentation to link to regarding how they use the `main` key, that would be valuable.

FDiskas:
Usually bundlers are resolving packages entry points specified in this `main` keyword
There is more discussion https://github.com/webpack/webpack/issues/1789#issuecomment-250964899
Documentation https://docs.npmjs.com/files/package.json#main

FDiskas:
I believe that this change will help others to use you lib with bundlers like webpack...

tysongach:
This is a bit of a brain dump because I don’t fully understand how all of the different JavaScript Sass implementations handle `@import` and how the `main` field works…

- This change—as is—would break our [eyeglass integration](https://github.com/thoughtbot/bourbon/blob/master/eyeglass-exports.js)
    - The entire premise of eyeglass, to me, is basically making `@import "bourbon"` work
    - Have you tried using eyeglass?
- This change also sort disregards entirely our `index.js` file in this repo and it is there to make `@import` work within node-sass
    - You can read about node-sass's `includePaths` here
    - So, I also think this change—as is—would break our node-sass integration
- It seems this change is based entirely around using [Dart Sass](https://github.com/sass/dart-sass) with Bourbon.
- I’m not sure just placing the path to our primary Sass manifest magically makes Sass `@import` work, as you seem to indicate
    - If you have documentation on this, please do share!
    - npm is primarily designed for JavaScript, and I think—generally speaking—most things expect `main` to point to a JavaScript file to run something
    - Long ago, we [had `main` set to our manifest](https://github.com/thoughtbot/bourbon/blob/76a61d9650ded0a79b4e08a28f327dc184ebabc1/package.json#L20) but Sass `@import` did not work then
- A [`style` field was added to our `package.json` in `c10f0bf`](https://github.com/thoughtbot/bourbon/commit/c10f0bf18feafa866a51259b65c765642c62bf8c#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) which points to `core/_bourbon.scss`, as you’ve done here
    - The impetus for that change and conversation in PR https://github.com/thoughtbot/bourbon/pull/978 was very similar to this one, which was desire to have `@import "bourbon"` work in their Node/npm-based workflows. Though, it should be noted that only _specific_ Sass tools support that field.
    - I’m sort of regretting merging that change because I now think `sass` is a better field than `style` because Bourbon is pure Sass, not CSS.
- A completely different approach could be to _move_ the `core` directly on package build (pre-publish to npm), therefore the import could be something like `bourbon/bourbon` instead of `bourbon/core/bourbon` (a step in the right direction, perhaps)
- I have a hunch the a more robust solution will require us defining specific Sass compiler support and somehow checking for which one is running when someone is trying to use Bourbon…
- I’m sort of wondering if we even need to support node-sass, given that Dart Sass is now the primary implementation
- I miss the old days with Ruby Sass was the thing, and `@import` just worked nicely in Rails apps.

Finally, if you can, please provide instructions for how to test this change. What are the exact steps you took to test this on your end?

tysongach:
I pulled down your [demo repo](https://github.com/FDiskas/parcel-bug) and manually made this change to Bourbon’s `package.json` inside `node_modules` and the issue still persists.

I’m going to close as I’m unable to find any documentation that this fix resolves the issue, and in testing am unable to confirm the fix, as well.

meduzen:
Side note here:

> I’m sort of wondering if we even need to support node-sass, given that Dart Sass is now the primary implementation

This makes me think about something Bourbon maintainers are maybe not aware of: `@import` is now deprecated and will be removed from SASS in < 2 years, replaced by a module system that’ll require from SASS libraries maintainers to update their code.

On this topic, I recommend:
- article: [_Introducing SASS modules_](https://css-tricks.com/introducing-sass-modules/)
- me [praising the module system but complaining about `@import` removal](https://github.com/sass/sass/issues/2750#issuecomment-541285541).

PR #1115, opened 2023-01-20
Update CI config ruby

saturnflyer:
## Description

Switch to Ruby 2.7.7 on CI


brimatteng:
Is it possible to also run this gem against any supported version of ruby? IS that a thing we want todo?

saturnflyer:
@brimatteng Yes. I want to update the config to run against a matrix

elisuh (approved):

PR #1118, opened 2023-01-23
Boubon v7.3.0

elisuh:
## Description

Bourbon v7.3.0


lib/bourbon/version.rb
```diff
@@ -1,3 +1,3 @@
 module Bourbon
-  VERSION = "7.2.0"
+  VERSION = "7.2.1"
```
brimatteng, line 2: This might want to be an 8.0.0 release or at least 7.3.0 since we are dropping support for older than rails 5 apps

elisuh: tysm for this, I have no idea how release numbers work!

brimatteng (approved):

PR #1133, opened 2024-09-13
Add note to readme about archiving and how to migrate

enatario:
❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 [![Reviewed by Hound](https://img.shields.io/badge/Reviewed_by-Hound-8E64B0.svg)](https://houndci.com)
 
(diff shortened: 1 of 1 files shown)


stefannibrasil (commented):
Very cool! thank you so much for your work on this!!

What do you think of putting this message right at the top? that has been the pattern I've seen for other projects. For example: https://github.com/thoughtbot/griddler?tab=readme-ov-file#deprecated-as-of-june-6-2024

enatario:
@stefannibrasil makes sense to me! Adjusted (still put it under the Repo name to maintain some hierarchy)

stefannibrasil (approved):
:shipit:

